PR #2, opened 2015-09-14
[Numerics] update CSparse

xhub:
The version of CSparse in Numerics is very old (2006). This commit updates it.

Nothing much to note, except:
- they change their indices to csi, which is defined to ptrdiff_t if we
  are not compiling to MATLAB. I kept the name csi when dealing with it
  in order to ease the transition if this gets renamed
- cs_lusol has its signature (order is now the first arg)
- the BIGGEST issue is in the css struct, where they use define lnz and
  unz as double, but use them as pointers in the code. By looking at the
  function cs_cumsum, it looks like they do that to prevent some
  overflow ... well the real fix should be to use size_t, but then the
  logic at the end of cs_schol is broken. I left them as csi (which is
  like we had before). I don't think we will ever overflow in our case.
  Feel free to give our opinion.

### DIFF
--- a/Numerics/src/NonSmoothSolvers/FrictionContact/GlobalFrictionContact3D_AlartCurnier.c
+++ b/Numerics/src/NonSmoothSolvers/FrictionContact/GlobalFrictionContact3D_AlartCurnier.c
@@ -63,7 +63,8 @@ unsigned int sizeOfPsi(
   CSparseMatrix* M,
   CSparseMatrix* H)
 {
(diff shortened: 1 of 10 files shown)


bremond:
conflicts + compil
https://github.com/bremond/siconos/tree/xhub-CSparse-update

xhub:
Thanks for the review and update. I pushed your branch to master directly to avoid 2 (meaningless merges)

PR #3, opened 2015-10-13
One cmake to rule them all

fperignon:
Change siconos distribution : from several independent components (Numerics, Kernel ...)
to one single package with a common cmake setup to build the library.


fperignon:
Merge "new cmake setup" into siconos master

PR #5, opened 2016-01-15
Fix qgl viewer examples

radarsat1:
Fix QGLViewer-based examples.  Notable changes:

  * Order of some operations in exported CMakeLists.txt was changed to take libraries added by some options into account.  Currently options WithQGLViewer and WithOpenGL specified in the example-specific .cmake file were ignored.
  * To avoid a segfault on Ubuntu 15.10, I found I needed to compile QGLViewer myself for Qt4, since the version installed by apt-get was compiled for Qt5.  For this, I added the Siconos install path to the paths searched in FindQGLViewer.cmake.
  * There was an out-of-bounds assertion in SphereLDS.cpp, related to qMemory being checked after initialize, but before anything is inserted.
  * Had to "#include" SiconosKernel.hpp before SiconosBodies.hpp, otherwise code would not compile, complaining about parameter mismatch for the visitor pattern.


radarsat1:
Ah, just realized this is based on 3.8.x.  Should I rebase it on master?

radarsat1:
Just rebased it on master, seems to make more sense. Sorry about that.

PR #9, opened 2016-01-28
Fix C++11 support in CMake and headers.

radarsat1:
It seems that compiling software that uses C++11 against Siconos would successfully compile, but not link.  It turned out be because, by default, Siconos compiles without -std=c++11, and uses boost::shared_ptr, etc.  However, the use of boost::shared_ptr or std::shared_ptr is detected by the preprocessor by checking against __cplusplus in some headers.  So software including those headers when -std=c++11 was enabled would incorrectly choose the std:: namespace, when Siconos was compiled with the boost:: namespace.

This patch series:
  * Adds USE_CXX11 to CMake, which requests -std=c++11 enabled, and checks that it works during configuration.  An error is emitted if it is requested but not supported.  It is disabled by default.
  * Fixes some problematic areas in mechanics where boost::array or similar was used even when c++11 was enabled.
  * Replaces checks against __cplusplus in the headers with checks against SICONOS_CXXVERSION, which is hardcoded at configure time, guaranteeing that user software will select the shared_ptr type that Siconos was compiled with.

### DIFF
--- a/cmake/cxxTools.cmake
+++ b/cmake/cxxTools.cmake
@@ -65,16 +65,42 @@ function(detect_cxx_standard output_var)
 
         enable_language(CXX)
 
-        try_run(
+        option(USE_CXX11 "Prefer C++11 features over BOOST whenever possible" OFF)
+
+        if(USE_CXX11)
+          try_run(
             run_result_unused
             compile_result_unused
             "${CMAKE_BINARY_DIR}"
             "${CMAKE_BINARY_DIR}/cxxversion.cpp"
+            COMPILE_DEFINITIONS "-std=c++11" # Try first with C++11 features enabled
             COMPILE_OUTPUT_VARIABLE CXXVERSION
-        )
+          )
+        endif(USE_CXX11)
 
         # Parse the architecture name from the compiler output
         string(REGEX MATCH "CXXVERSION ([a-zA-Z0-9_]+)" CXXVERSION "${CXXVERSION}")
+
+        if(NOT CXXVERSION)
+          if(USE_CXX11)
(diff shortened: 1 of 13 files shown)


xhub:
Good idea!

radarsat1:
Okay I have been looking a little deeper into CMake support for C++11, and here's a slightly different proposal:

  * I tried to "do the right thing" and use CMAKE_CXX_STANDARD and module WriteCompilerDetectionHeader along with target_compile_features() for detecting and specifying specific C++ features.  CMake is then responsible for determining the C++ standard to use and how to use it. 
  * However, I discovered that mysteriously CMake doesn't provide any feature named "cxx_memory_shared_ptr" or similar -- so it's a bit of a dead-end to let CMake take care of this.

However, the idea of detecting specific features is good.  So, I propose:

  * User can set CMAKE_CXX_STANDARD and USE_BOOST_FOR_CXX11 to whatever he wants.
    - Therefore we presume he knows that enabling CMAKE_CXX_STANDARD=11 will enable namespace SP=std, unless USE_BOOST_FOR_CXX11 is enabled.
  * If USE_BOOST_FOR_CXX11 not set, check if std::shared_ptr is supported.
    - If yes, use namespace SP=std and set a proprocessor variable SICONOS_STD_SHARED_PTR
    - Otherwise, do not use it.
  * We mostly forget about SICONOS_CXXVERSION, and use feature variables in all cases.
  * Variables SICONOS_STD_SHARED_PTR and SICONOS_USE_BOOST_FOR_CXX11 are recorded to SiconosConfig.h and used in headers

PR #10, opened 2016-02-01
[cmake] Add detection of C++1x features, and store results in SiconosConfig.h

radarsat1:
This PR includes the changes from PR #9, but proposes a different approach for handling C++1x features.  To summarize:

> The use of boost::shared_ptr or std::shared_ptr is detected by the preprocessor by checking against __cplusplus in some headers. So software including those headers when -std=c++11 was enabled would incorrectly choose the std:: namespace, when Siconos was compiled with the boost:: namespace.  This caused linker errors.

This patch:

  * Adds detection of individual C++11 features, such as std::shared_ptr, std::array, std::unordered_map, std::to_string, std::bind.
  * Detection variables are stored in SiconosConfig.h
    - e.g., SICONOS_STD_SHARED_PTR, etc..
  * User flags to turn *off* use of these features are also stored in SiconosConfig.h:
    - SICONOS_USE_BOOST_FOR_CXX11
    - SICONOS_USE_MAP_FOR_HASH
  * Fixes some problematic areas in mechanics where boost::array or similar was used even when c++11 was enabled.
  * Replaces checks against __cplusplus in the headers with checks against specific feature variables.

Note that CMake's own mechanism for setting the C++ standard must be used to select C++11 or C++14:

    cmake .. -DCMAKE_CXX_STANDARD=11

This option is explicitly passed on to the detection try_run() calls.

### DIFF
--- a/io/src/Register.hpp
+++ b/io/src/Register.hpp
@@ -28,7 +28,7 @@
 #include <boost/serialization/serialization.hpp>
 #include <boost/serialization/nvp.hpp>
 
-#if __cplusplus >= 201103L
+#if defined(SICONOS_STD_SHARED_PTR) && !defined(SICONOS_USE_BOOST_FOR_CXX11)
 #include <boost/serialization/ser_shared_ptr.hpp>
 #else
 #include <boost/serialization/shared_ptr.hpp>
--- a/wrap/swig/sharedPointers.i
+++ b/wrap/swig/sharedPointers.i
@@ -7,7 +7,7 @@
 
 %import SiconosConfig.h
 
-#if (SICONOS_CXXVERSION >= 201103L) && !defined(USE_BOOST_FOR_CXX11)
+#if defined(SICONOS_STD_SHARED_PTR) && !defined(SICONOS_USE_BOOST_FOR_CXX11)
 #define STD11 std
 #undef __cplusplus
 #define __cplusplus SICONOS_CXXVERSION
--- a/cmake/CXXCompilerSetup.cmake
+++ b/cmake/CXXCompilerSetup.cmake
@@ -10,10 +10,30 @@
 if(NOT CMAKE_CXX_COMPILER OR NOT CMAKE_CXX_COMPILER_WORKS)
   message(ABORT "no cxx compiler")
 endif()
-  
+
+# For SiconosConfig.h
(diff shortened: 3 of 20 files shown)


io/src/Register.hpp
```diff
@@ -28,7 +28,7 @@
 #include <boost/serialization/serialization.hpp>
 #include <boost/serialization/nvp.hpp>
 
-#if __cplusplus >= 201103L
+#if SICONOS_CXXVERSION >= 201103L
```
radarsat1: Note: This was the only one I wasn't sure to do with in terms of feature detection, so I at least changed the comparison with __cplusplus for SICONOS_CXXVERSION.

wrap/swig/sharedPointers.i
```diff
@@ -7,7 +7,7 @@
 
 %import SiconosConfig.h
 
-#if (SICONOS_CXXVERSION >= 201103L) && !defined(USE_BOOST_FOR_CXX11)
+#if defined(SICONOS_STD_TO_STRING) && !defined(SICONOS_USE_BOOST_FOR_CXX11)
 #define STD11 std
```
xhub: I think you meant SICONOS_STD_SHARED_PTR: the following bits of code are from g++-v4/bits/shared_ptr.h

io/src/Register.hpp
xhub: I did this 3.5 years ago (thanks git blame for reminding me this), so I had to take a stab at it and try to remember ... So looking at the code, ser_shared_ptr.hpp does the serialization of the std shared_ptr. Hence, I would add 
#if defined(SICONOS_STD_SHARED_PTR) && !defined(SICONOS_USE_BOOST_FOR_CXX11)
here.

xhub:
I like this new version. I just had 2 minors comments. If you think the proposed changes make sense, just commit an updated version to the main repo.

wrap/swig/sharedPointers.i
radarsat1: Right, good catch!  Must have done that one a little too quickly.

radarsat1:
Ok I updated the branch but apparently I don't have push access for the siconos repository.

xhub:
I added you to the siconos group. I think that's the way to go.

radarsat1:
Seems to have worked, thanks!

PR #75, opened 2016-06-20
Use OpenMP in fc3d_nsgs().

radarsat1:
Please review!  I am still testing.


radarsat1:
I (maybe not very intelligently) rebased this on top of master before pushing it, and now it breaks.  Something with the recent initialization changes.  Will check tomorrow.

radarsat1:
No it builds, sorry, I just hadn't cleaned my swig directories.

radarsat1:
figure_1

Results from running chute.py on 12-core computer with different values of `OMP_NUM_THREADS`.

vacary:
Good job.

Really interesting and encouraging.

Up to 6 threads, the gain is really impressive. After, It is not so clear but are you sure that you have a sufficient number of contact.  I am wondering if it will not be interesting to start to parallelize after, let us say, 20 iterations, to avoid the overload due to OPENMP on problems that are "easy"

radarsat1:
That's right, to fully characterize we'll need to test on a variety of problem types (number of objects, number of contacts, etc.)
Agreed on delayed parallelization, or having some conditions (time, problem size) that should be met.
Anyways, it's a start :)  I haven't tested against reference files for all existing tests however, this should be done before merging.

PR #77, opened 2016-06-23
[cmake] Use ccache if available.

radarsat1:
I find this very useful, have been using it for months, thought I may as well propose it for everyone..

### DIFF
--- a/cmake/SiconosSetup.cmake
+++ b/cmake/SiconosSetup.cmake
@@ -204,3 +204,9 @@ if(WITH_DOCUMENTATION)
   set(USE_SPHINX TRUE)
 endif()
 
+# =========== use ccache if available ===========
+find_program(CCACHE_FOUND ccache)
+if(CCACHE_FOUND)
+  set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
+  set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ccache)
+endif(CCACHE_FOUND)


radarsat1:
For the record, I'm making PRs to check things on Travis and get opinions from everyone, but I'm happy to rebase and commit to master myself for these kind of small things after some approval/discussion.

xhub:
Just FYI I've been an avid user of ccache back in the day. But it can also induce subtle bugs, especially if you switch compilers to do your developement. I'd add a switch to protect from the unconditional use.

vacary:
I suggest to merge this pull request.

radarsat1:
Meant to get around to adding it as optional.  I'll do so now.  Only question is, should it be default on or default off?

xhub:
I'd vote for default off. It can introduce subtle bugs (if the compiler is updated, the cache should be cleaned) and if it is the default, this may create problem for users.

I could not test the following, but why do we need code in the cmake files to use ccache? Settings CC and CXX to ccache should do the trick no?

Also there is another option mentioned here:
http://stackoverflow.com/questions/1815688/how-to-use-ccache-with-cmake#1954144

And changing RULE_LAUNCH_LINK  has no benefit, see
https://github.com/facebook/hhvm/commit/b1bed34feb35138743ff163bc2b731f502d0f50e

IFAIK, ccache does not improve the link step.

radarsat1:
Yeah, ok I turned it off by default.  It's a good point that it's pretty easy to add via the compiler options, but anyway, doesn't hurt to keep it I guess, it's nice to have the detection, so I think this is a good compromise.

PR #79, opened 2016-07-13
Fix build error with non MSVC compiler on Windows platform

moleculext:
I've tried to build siconos on Windows inside MSYS2. MSYS2 uses MinGW as compiler and I've encountered errors about the `/bigobj` command option.

This option is only for the MSVC compiler, hence the fix that replaces `CMAKE_SYSTEM_NAME MATCHES Windows` by `MSVC` variable in cmake files.

I can't test the fix with the Visual Studio Compiler, but it works within the MSYS2 platform using MinGW-32 and MinGW-64.

Thanks,
François

### DIFF
--- a/cmake/PythonWindowsSetup.cmake
+++ b/cmake/PythonWindowsSetup.cmake
@@ -8,7 +8,7 @@ if(WITH_TESTING)
   ENDIF()
 endif()
 
-IF(CMAKE_SYSTEM_NAME MATCHES Windows)
+IF(MSVC)
   SET(VAR_SEPARATOR ";")
   APPEND_CXX_FLAGS("/bigobj")
 ELSE()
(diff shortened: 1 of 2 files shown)


xhub:
Thanks for the fix. A few years back I experimented with msys, but I had numerical issues: many simulation results where quite different. When I switched to Visual Studio, those issues disappeared.

Did you run the examples? If yes, where the results close enough to the reference ones?

radarsat1:
Really! Surprising, but possible I guess.  No windows computer here that I can help test on.  One thing I've done in the past with MingW is use the Ubuntu packages to compile the Windows version on Linux, and test using Wine.  A bit convoluted, but actually quite practical.

xhub:
MSYS does not use glibc as libc, so at the time my hypothesis was that this was the most likely culprit. But this may have changed recently.

With the new linux subsystem on windows, maybe we can achieve something nicer.

radarsat1:
Yep, it's pretty much worth waiting just for that ;)

PR #81, opened 2016-08-12
[numerics] separate out solver steps and collapse into one while loop

radarsat1:
Has one extra while loop dedicated to the most common case, but dispenses with having a different loop for every possible combination of options.

Wanted to check at least with @vacary before pushing this one since it might affect current work.  I thought it might help to clean up all those separate loops and avoid having to copy-paste changes all over the file, but let me know if you think it's a bad idea..

### DIFF
--- a/numerics/src/FrictionContact/fc3d_nsgs.c
+++ b/numerics/src/FrictionContact/fc3d_nsgs.c
@@ -70,9 +70,14 @@ void fc3d_nsgs_update(int contact, FrictionContactProblem* problem, FrictionCont
 
 
 }
-void fc3d_nsgs_initialize_local_solver(SolverPtr* solve, UpdatePtr* update, FreeSolverNSGSPtr* freeSolver, ComputeErrorPtr* computeError,
-                                FrictionContactProblem* problem, FrictionContactProblem* localproblem,
-                                SolverOptions * options, SolverOptions * localsolver_options)
+
+void fc3d_nsgs_initialize_local_solver(SolverPtr* solve, UpdatePtr* update,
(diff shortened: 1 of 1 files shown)


radarsat1:
ps. if accepted i'll see about porting the openmp branch.. hopefully doesn't cause too much trouble there, but I suppose the openmp `for` loops would have to remain a separate case.

vacary:
In fact, I was the author of the splitting to avoid the use of a sensible of switches, conditional statements and declarations on the stack and on the heap.
@bremond told me several times that it was useless to multiply the cases.
If you are sure that the cost of doing the conditional statements in each loop rather than one time is the same or taken into account by optimization. let us merge.

radarsat1:
Actually I would say the patch has two goals:
1. To break the loop up into small pieces so that it is easily understandable and avoids code copying, but still reconfigurable.
  * A result of the above is that we can compress each `while` into smaller space, my goal was to be able to see the whole logic in one screen.
  * I used static functions purposely to make it dead easy for the compiler to inline things.
2. Secondly, we can still specialise for specific cases as before.  However, I prefer to have all the booleans in one place at the top of each case instead of multiple nested if statements, as I think it's much more clear.
  * The remaining cases are still taken care of by embedded `if` statements, which honestly I think have minimal impact for rare cases.  Otherwise there is really a combinatorial explosion of specialisation cases, without C++ templates this is difficult to handle succinctly.

Anyways, mostly trying to reduce space for making mistakes..

All open to opinion of course!

radarsat1:
Rebased in c83e92f12f5b69f9d2c2b1b4275c44eee45f0178

PR #86, opened 2016-10-25
Mechanics proposed

radarsat1:
I am pretty confident now that all examples work as expected.  The new Mechanics API allows a mechanics_io-like interaction on the C++ level (i.e. declaration of abstract shapes implemented by a hidden Bullet-based back-end), and integrates this into mechanics_io.py.  It gives per-shape control over collision margins and a few Bullet-specific options such as gContactBreakingThreshold and support for scaling in the collision engine.

Please review.

Adds a backend selection interface, instead of relying only on overriding classes.  This (and the backend options) is not saved i the HDF5 file. Overriding classes is still supported, although there was an issue with different constructor arguments for the TimeStepping override which has the potential to lead to a bit of confusion.  Tried to make the OCC example work to be sure not to break things completely but it seems still in need of some work in other areas.

In any case, now that examples are working, I feel confident that further work could proceed on the master branch.


radarsat1:
Fixed OCC example.

PR #99, opened 2017-01-19
Lazy initialization

vacary:
Hi,
I started to  work towards some lazy initialization of members in the modeling objects. I have done a first implementation for NewtonEulerDS and LagrangianDS where the initialize method is now empty.

The objectives are:
- to initialize the attributes of modeling objects if we need it only. 
- to initialize the object even if the simulation initialize has not been called. The goal is to be able to use modeling objects as DynamicalSystem in a simulation written  in Python independently of the simulation in Siconos
- that the simulation initializes only "simulation attributes" in modeling objects (memory, reset initial state, ...)

The main changes for users is when we derive a class, we need to initialize memory of the object that we want to use. We can no longer rely on a big initialize when the simulation  starts. The will have some effects on the examples and modules where we derive classes. Let us see.

To avoid to test it at each calls of computeSomething, I put the allocation in setPluginSomething. We can got further and put the allocation in computeSomething. What do you think ?

I leave review the modifications and let me know what you think. If ok, I will continue on FirstOrderDS and then to Interactions.

### DIFF
--- a/kernel/src/modelingTools/DynamicalSystem.cpp
+++ b/kernel/src/modelingTools/DynamicalSystem.cpp
@@ -118,6 +118,19 @@ void DynamicalSystem::zeroPlugin()
   _pluginJacxDotG.reset(new PluggedObject());
   _pluging.reset(new PluggedObject());
 }
+
+void DynamicalSystem::resetAtInitialState()
+{
+  if(_x0)
+  {
+    *(_x[0]) = *_x0;
+  }
+  else
+    RuntimeException::selfThrow("DynamicalSystem::resetAtInitialState() - initial state _x0 is null");
+  
+}
+
+
 // Setters
 
 void DynamicalSystem::setX0(const SiconosVector& newValue)
--- a/kernel/src/modelingTools/DynamicalSystem.cpp
+++ b/kernel/src/modelingTools/DynamicalSystem.cpp
@@ -119,14 +119,14 @@ void DynamicalSystem::zeroPlugin()
   _pluging.reset(new PluggedObject());
 }
 
-void DynamicalSystem::resetAtInitialState()
+void DynamicalSystem::resetToInitialState()
 {
   if(_x0)
   {
     *(_x[0]) = *_x0;
   }
   else
-    RuntimeException::selfThrow("DynamicalSystem::resetAtInitialState() - initial state _x0 is null");
+    RuntimeException::selfThrow("DynamicalSystem::resetToInitialState() - initial state _x0 is null");
   
 }
 
--- a/kernel/src/modelingTools/DynamicalSystem.hpp
+++ b/kernel/src/modelingTools/DynamicalSystem.hpp
@@ -255,6 +255,10 @@ public:
    */
   virtual bool checkDynamicalSystem() = 0;
 
+  /** reset the state to the initial state */
+  virtual void resetAtInitialState();
+
+  
   /*! @name Members access */
   //@{
 
--- a/kernel/src/modelingTools/DynamicalSystem.hpp
+++ b/kernel/src/modelingTools/DynamicalSystem.hpp
@@ -256,7 +256,7 @@ public:
   virtual bool checkDynamicalSystem() = 0;
 
   /** reset the state to the initial state */
-  virtual void resetAtInitialState();
+  virtual void resetToInitialState();
 
   
   /*! @name Members access */
--- a/kernel/src/modelingTools/FirstOrderNonLinearDS.cpp
+++ b/kernel/src/modelingTools/FirstOrderNonLinearDS.cpp
@@ -208,8 +208,6 @@ void FirstOrderNonLinearDS::initialize(double time, unsigned int sizeOfMemory)
   if (! _z)
     _z.reset(new SiconosVector(1));
 
-  // Initialize memory vectors
-  initMemory(sizeOfMemory);
 
   updatePlugins(time);
   if (_f)
--- a/kernel/src/modelingTools/LagrangianDS.cpp
+++ b/kernel/src/modelingTools/LagrangianDS.cpp
@@ -26,7 +26,7 @@
 
 
 // Private function to set linked with members of Dynamical top class
-void LagrangianDS::connectToDS()
+void LagrangianDS::connectToDS(unsigned int steps)
 {
   // dim
   _n = 2 * _ndof;
@@ -43,6 +43,7 @@ void LagrangianDS::connectToDS()
   _x[0].reset(new SiconosVector(*_q[0], *_q[1]));
   _x[1].reset(new SiconosVector(*_q[1], *_q[2]));
   // Everything concerning rhs and its jacobian is handled in initRhs and computeXXX related functions.
+  DynamicalSystem::initMemory(steps);
 }
 void LagrangianDS::zeroPlugin()
 {
@@ -58,7 +59,8 @@ void LagrangianDS::zeroPlugin()
 }
 
 LagrangianDS::LagrangianDS(SP::SiconosVector newQ0, SP::SiconosVector newVelocity0):
-  DynamicalSystem(2 * newQ0->size()), _ndof(newQ0->size())
+  DynamicalSystem(2 * newQ0->size()), _ndof(newQ0->size()),
+  _hasConstantFExt(false)
 {
   zeroPlugin();
   // -- Memory allocation for vector and matrix members --
@@ -77,8 +79,12 @@ LagrangianDS::LagrangianDS(SP::SiconosVector newQ0, SP::SiconosVector newVelocit
   //   _xq.reset(new SiconosVector(dimension()));
   //   mXfree.reset(new SiconosVector(dimension()));
   //   r.reset(new SiconosVector(dimension()));
-
   /** \todo lazy Memory allocation */
+  _forces.reset(new SiconosVector(_ndof));
+  _jacobianqForces.reset(new SimpleMatrix(_ndof, _ndof));
+  _jacobianqDotForces.reset(new SimpleMatrix(_ndof, _ndof));
+
+   /** \todo lazy Memory allocation */
   _p.resize(3);
   _p[0].reset(new SiconosVector(_ndof));
   _p[1].reset(new SiconosVector(_ndof));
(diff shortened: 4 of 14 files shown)


radarsat1:
I think getting rid of as many explicit 'initialize' calls as possible is the way to go. I was struggling with a problem just this week where I could not figure out why the level set was incorrect until I realized that even though I called NSDS::link(), I forgot to call Simulation::initializeInteraction().  So this is a good start.

On the other hand, localizing all the memory handling in initialize() provides an advantage for code organization.  Instead of moving all the "new" around, why not just put the following at the top of each function?

    if (! is_initialized) initialize();

This would have the advantage of being one single very fast test instead of having to test a null pointer for each vector.  The pointer checks could be changed for debug asserts -- we could make a macro,

    USES(_p[0], ndof);

that declares that a function uses some vector of an expected size and translates into,

    if (!_initialized) { initialize(); _initialized=true; } assert(_p[0] && _p[0]->size() == ndof);

where the assert() goes away in debug build.

This kind of thing I think is a bit dangerous,
~~~
if (_x[0] && _xMemory)
    _xMemory->swap(*_x[0]);
~~~
this should really be an assertion, no?  Is there a case where _xMemory() can be valid as null?

Otherwise you could get hard-to-detect numerical errors when you think memory has been swapped but it has not because `_x[0]` is allocated but `_xMemory` is not.

** English note, it would be "resetToInitialState" rather than "resetAtInitialState".

vacary:
Concerning
`if (! is_initialized) initialize();`
I understand that a simple initialization that collects all memory allocation is easier to maintain and implement. But I see major drawbacks:
- such an implementation is difficult to implement with a minimal memory cost
- its effect will depend when we call it. I want to avoid such a 'black box' effect for the user.
- its is difficult to explain to user why we need to initialize the objects of the modeling classes it we want to use into another simulation software, or just doing some simulation scripts in Python.

Concerning
`USES(_p[0], nodo);`
I agree that such a macro is useful. It is very similar to some concepts of [design by contract] (https://en.wikipedia.org/wiki/Design_by_contract). But is it restricted to debug mode and if it is not allocated nothing will be done to fix the issue.

radarsat1:
Ok, thanks for the explanations on all points.  To continue clarification,

> its effect will depend when we call it. I want to avoid such a 'black box' effect for the user.

I was proposing having this at the top of any function that uses a vector variable.  The user (the caller) would not need to know about the call to initialize(), it would be internal to the object.  But I realize maybe I am not thinking about the word "user" the same way -- do you mean "user" as in derived classes?  In that case yes, they would need to know.

> But is it restricted to debug mode and if it is not allocated nothing will be done to fix the issue. I wanted to get a self-adaptive behavior if we need _p[0] but it is not allocated.

I was proposing that the USES() macro would, in debug mode, allocate _p[0] if it is not there, and check that the allocation was successful and the size is correct, while in release mode, it would allocate but not check.  Anyway, just an idea.

> the LagragianDS class must work without _x[0] and _xMemory

Fair enough.  I guess if _xMemory does not exist we consider it equivalent to a SiconosMemory of size 0, which is anyway ignored by the integrator.  In reality this is memory for the integrator rather than for the DS, right?

In a way it would be much simpler to have static SiconosMemory objects in the class, with variable size, instead of dealing with nullable pointers to them at all.  (After all they contain pointers to deque objects that contain an array of pointers to SiconosVector objects that contain a pointer to a ublas vector.  That's quite a lot of pointer indirection already.)

bremond:
I think that an eventual solution in the case of values which may be
nothing or something (as 'a option, None or Some _ in Ocaml) is
boost::optional

It can be considered simply equivalent to:

template <typename T>
class optional
{
  bool _initialized;
  T _storage;
};

with some details on memory alignment:
http://www.boost.org/doc/libs/1_63_0/libs/optional/doc/html/boost_optional/tutorial/performance_considerations.html

Also, it seems thread safe

http://stackoverflow.com/questions/20271300/is-it-safe-to-use-boostoptional-in-interprocess-memory

And from the boost documentation:

"It is recommended to use optional<T> in situations where there is
exactly one, clear (to all parties) reason for having no value of type
T, and where the lack of value is as natural as having any regular value
of T"


with boost::optional<SomeC> x = SomeC()

One can:

 - check if there is some value
   if (x) { ... }  // if x is not None 

 - get the value
   *x or x.value() (the get() method seems to have been dropped by
   the c++17 standard)

 - get the value if there is one or give a default
   x.value_or(some_value)


With the latest some lazy initialization may be done like this:

#include <iostream>
#include <boost/optional/optional.hpp>

int main()
{
  boost::optional<int> o = boost::none;

  int i = o.value_or((o=1, o.value())) + 1;

  std::cout << i;
}

Furthermore, as there is not the overhead of reference counting, an
optional access must be more efficient than a shared pointer access and
probably in many places in the kernel where a pointer is needed only for
some delayed initialization this would be a benefit.

For the standard, std::optional exists in c++17

http://en.cppreference.com/w/cpp/utility/optional

(and the cppreference is far more usable than boost doc...)

>In a way it would be much simpler to have static SiconosMemory objects in the class, with variable size, instead of
>dealing with nullable pointers to them at all. (After all they contain pointers to deque objects that contain an
>array of pointers to SiconosVector objects that contain a pointer to a ublas vector. That's quite a lot of pointer
>indirection already.)

To be checked, but the nullable pointers may be some boost::optional
and we avoid a pointer indirection ?

(I do not agree with your static SiconosMemory solution. Even if
everything is sequential today, I think static members (and any kind of
global state) should be avoided as it would not facilitate any attempt
to use some parallelism in the kernel.)

radarsat1:
Hm, I'm not sure what it may have to do with parallelism, but I agree that optional can be a nice way to document things that are really purposely "optional" as opposed to just "accidentally nullable" like a pointer.  You're right that there is the benefit of skipping the reference count too, although we'd get it back if `std::optional<SP::SiconosMemory>` were used.

Anyways, on the pull request, concerning

> if (_x[0] && _xMemory) _xMemory->swap(*_x[0]);
> I recognized that it is a bit dangerous. But the LagragianDS class must work without _x[0] and _xMemory

Ok I understand that basically we only want to push a value into `_xMemory` if it exists.  If `_x[0]` does not exist, then we ignore both.  So the code is correct, I understand it better now, and I'm ok with the pull request then.  I realize the USES idea is difficult since the DS does not have enough information to make the decision about whether a variable is needed or not, as it depends on the integrator, not on the DS.

I still think something more optimal could be done with the indirections here but that's for another topic.

radarsat1:
I compiled this and ran a few examples without problems.  Merge?

vacary:
I will add a last commit and we should be able to merge.

vacary:
Perhaps, we should translate the discussion to an issue to continue to discuss this point ?

What do you think ?

bremond:
Maybe several issues with what Steve said

 - static memory for work vectors ==> this reminds me the question about
   the usage of memory pools (I do think there is no emergency here in
   front of solvers performance, unless it can be stated that everything
   really suffers from accesses to those different allocated places)

 - multiple pointers indirections in SiconosVector ==> the need for
   static size siconosVector, SiconosVector3, SiconosVector6 based on
   c++11 std::array, (with no pointer on it!). This would be mainly for
   NewtonEuler usage and, in my sense, this is more urgent than memory
   pools.

 - lazy initialization ==> is std::optional a good base for this ?
   (Although a shared pointer can do the job, I think it is worth to ask
   the question because it can clarify the usage of class members and
   then let us understand what we are doing and where heavier shared
   pointers are really needed. Also, this would imply to review shared
   pointers members that may be transformed to simple optional values.)
   The question about when to do it remains...

bremond:
>Hm, I'm not sure what it may have to do with parallelism,

In fact many solutions to lazy init in c++ lead to some form of the
singleton pattern which uses some static flag, I wanted to check that it
is not the case with boost::optional

>You're right that there is the benefit of skipping the reference count too, although we'd get it back if std::optional
><SP::SiconosMemory> were used.

I mean to get rid off the SP:: thing and use directly
std::optional<SiconosMemory>, and so on

in that case this code

if (_x[0] && _xMemory) _xMemory->swap(*_x[0]);

remains unchanged.

This can be seen as only a technical detail, we just remove the
reference counting stuff, but as you said : this is also a nice way to
document things that are really purposely "optional". I think it is
fundamental...

radarsat1:
Yes, but be aware that  `sizeof(std::optional<SiconosMemory>) ~= sizeof(SiconosMemory) + 8`, whereas `sizeof(std::optional<SP::SiconosMemory>) == sizeof(shared_ptr) + 8`.
That is, memory-wise it is the same as putting a SiconosMemory directly in your data structure.  In that case then, there is no benefit to not just putting a SiconosMemory statically into the structure and allowing it to have size 0.

i.e., compare (a)
~~~
struct DS {
   SiconosMemory _xMemory;
   std::shared_ptr<SiconosVector> _x;
};

DS ds;
ds._xMemory.swap(ds._x);
~~~

where swap() is robust to the condition that `_x==0` or `_xMemory.size()==0`

as opposed to, (b)
~~~
struct DS {
   std::optional<SiconosMemory> _xMemory;
   std::shared_ptr<SiconosVector> _x;
};

DS ds;
if (ds._x && ds._xMemory)
  ds._xMemory.swap(_x));
~~~
which provides little syntactical benefit over, (c)
~~~
struct DS {
   std::shared_ptr<SiconosMemory> _xMemory;
   std::shared_ptr<SiconosVector> _x;
};

DS ds;
if (ds._x && ds._xMemory)
  ds._xMemory.swap(_x));
~~~
and indeed takes up more memory in the case that `_xMemory.size()==0`.  I see an advantage for (a) (simpler syntax avoiding the null check, less indirection) or for the current case (c) (better memory efficiency, don't allocate SiconosMemory unless needed), but (b) seems to have the disadvantages of both.

bremond:
Yes, you are right. Anyway unlike with shared pointers, there is no
memory management with std::optional so it is only a way to wrap plain
objects. It is possible to wrap references also, but it only delays the
problem of creation and management. Finally, in these cases where no other
pointers to the data are needed, to avoid the useless reference
counting, one may (must?) use std::unique_ptr (since c++11).

so (d)

#include <memory>

struct DS {
   std::unique_ptr<SiconosMemory> _xMemory;
   std::shared_ptr<SiconosVector> _x;
};

DS ds;
if (ds._x && ds._xMemory)
  ds._xMemory.swap(_x));

is like (c) without the cost of refcounting for _xMemory access.

PR #100, opened 2017-01-26
Embed SiconosMemory in DS

radarsat1:
Here's a view of how the OSI / DS code could look if two changes are made:

1. Instead of SiconosMemory being a class containing a pointer to std::deque\<SP::SiconosVector\>, it now *is* a std::vector\<SiconosVector\>.  (Actually I couldn't think of a reason to keep it as deque..)
2. Instead of DS having SP::SiconosMemory, it has directly a SiconosMemory, which may have size 0.

Thus, instead of nullable shared pointers everywhere that need to be checked and reference counted, each DS just directly owns arrays of SiconosVectors for each memory.  To better model the fact that this data represents immutable history, SiconosMemory::getSiconosVector() now returns a `const SiconosVector&`.

(There were *two* places that to my surprised actually modify the contents of a SiconosMemory vector, so I was forced to add getSiconosVectorMutable() which does some ugly tricks.. I am not sure if those "memory overwrites" are necessary.)

Since in the OSI classes, many many vector accesses just deference the SP::SiconosVector temporaries, I took the liberty of changing as many of them as possible to SiconosVector& or const SiconosVector& as appropriate.  This takes a lot of extraneous \*s out of the mathematics, and avoids reference counting overhead for temporary variables.

(e.g. `*residuFree -= 0.5 * h**accFree` becomes `residuFree -= 0.5 * h*accFree`)

Places where I could not do this were generally whenever the SP::SiconosVector was subsequently passed to a function expecting one, such as `computeForces`.

Does it make a difference in efficiency?  I am not sure, I tested a few scripts and didn't find *increases* in timing.  I did find that the executation of n_cubes.py went from 74s to 55s on my computer, although other examples showed much less difference.  I am sure these reference counts are not the bottleneck.  The removal of SP::SiconosMemory puts the memory usage of SiconosMemory directly into the DS, so it does add some bytes to total memory usage (adds 24 bytes) but only in the case of an OSI that would leave them at size 0.  Otherwise it uses the same memory but with two fewer dereferences per access!

More important than the decrease in derefences and reference counting, I think that the use of references over SPs better reflects the semantics of the code and makes it easier to read, and the lack of nullable pointers makes for less explicit pointer validity checking and easier reasoning about memory, but this is of course just my opinion.  Open to discussion!


bremond:
>Since in the OSI classes, many many vector accesses just deference the SP::SiconosVector temporaries, I took the
>liberty of changing as many of them as possible to SiconosVector& or const SiconosVector& as appropriate. This
>takes a lot of extraneous *s out of the mathematics, and avoids reference counting overhead for temporary
>variables.

Yes, and this is important for code readability. To be more precise in

*residuFree -= 0.5 * h**accFree

the cost of reference counting is in the useless memory involved to
store the temporary shared pointers. The time of execution is not
affected, I think, because, the access is a dereference and for shared
pointers this is a simple data access in a structure, which should not
be more costly than a raw pointer access (as long as the structures does
not fill the L1 cache!)

After their creations, the time of execution cost of shared pointers is
present when there is a copy and probably the major drawback in siconos
is in argument passing of functions where there are plenty of them! This
is where reference counting is really in action. In those cases, if the
shared pointer cannot be avoided (API design) and if the time of
execution matters it is also possible to try to use some const
shared_ptr<const T> & (note that there is a SPC:: namespace for
shared_ptr<const T> &)

vacary:
Hi, I have changed a lot of things in my last commit and I continue to change a lot of stuff. For instance, I use more and more references.

radarsat1:
Yes, I will leave this PR here for a while (open for now) until you are finished your latest changes, then I will rebase it and consider whether it needs to be re-proposed.

radarsat1:
I have updated this PR to the current master and tested it.
Not sure if it's worth it, but I still think it's interesting to remove two levels of indirection at each access of memory values.

Issue #104, opened 2017-02-03
[swig] make sequenceToUnsignedIntVector support numpy array with int types


xhub:
Right now we expect an argument of the type std::vector<int> to be a sequence of int. We should support numpy arrays with type np.int{32,64,}

PR #105, opened 2017-02-03
[io] make wheel.py runnable

xhub:
This PR is not intented for merge, but to rather show the changes I had to make in order to make wheel.py run on my setup. I use python3, this may be the most noticible thing.

The list issue could be fixed when #104 is taken care of.

For the body_class, i think the best is to change the number of arguments rather than define a dummy one. I don't see how to fix that one properly since in C++ we have a default argument and that swig wrap both constructors.

### DIFF
--- a/io/swig/io/mechanics_io.py
+++ b/io/swig/io/mechanics_io.py
@@ -1089,7 +1089,7 @@ class Hdf5():
                     # TODO: segfault if we pass None for SiconosVector!
                     shp = self._shape.get(c.data)
                     pos = list(c.translation) + list(c.orientation)
-                    cset.append(SiconosContactor(shp, pos, c.group))
+                    cset.append(SiconosContactor(shp, pos, int(c.group)))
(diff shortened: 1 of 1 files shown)


radarsat1:
I'll compile for Python3 and test the examples shortly.

radarsat1:
With 077e2ef349 these problems are now fixed at the bindings level.  Any Siconos function accepting float or int should now properly convert numpy values, and functions accepting pointers to vectors or matrices should be able to accept None values which convert to null pointers.

My argument for the latter is that some functions *may* want to support optional values by means of accepting null, and previously this was rejected by the swig bindings.  Also, even if this could lead to segfaults, otherwise it would lead to cryptic "wrong arguments" error messages from the swig bindings.  I find it's actually easier to debug null-pointer segfaults in real Siconos code using gdb than understanding why swig is refusing to even call the function.

Functions with reference arguments (SiconosVector&) will give an error message, e.g. 'expected vector'.

Issue #119, opened 2017-03-29
How to initialize DS now, after recent refactor?


radarsat1:
I am adapting my code to the latest refactor, where a lot of initialization functions have changed.  I am having to make some guesses based on the diff, it would be nice to get an overview of the changes.  In any case, mostly it is clear, but I when initializing a new dynamically-added DynamicalSystem, I have a problem that I can't resolve so clearly.  I removed the call to DynamicalSystem::initialize(), since it seems that was removed.  However, now it crashes during Topology::initDS() because the workVector pointer is null.

This is the complete sequence I am using to initialize a new DS:

* model->nonSmoothDynamicalSystem()->insertDynamicalSystem(ds);
* model->nonSmoothDynamicalSystem()->topology()->setOSI(ds, osi);
* model->nonSmoothDynamicalSystem()->topology()->initDS(model, nextTime, ds, osi);
* ds->initialize(nextTime, osi->getSizeMem())  **removed**
* simulation->initialize(model, false)

Previously this worked, but now there is a segfault in MoreauJeanOSI::initializeDynamicalSystem(), called from Topology::initDS(), due to workVector==NULL.  (*_dynamicalSystemsGraph->properties(dsv).workVectors)  What step am I missing to initialize it?

radarsat1:
Ok I found that the initialization of workVectors was commented-out in Topology::insertDynamicalSystem, so assuming this is on purpose, I moved the initialization to where it is needed in MoreauJeanOSI::initializeDynamicalSystem().  Probably this then needs to be done in every OSI, or we should put it in a function in OneStepIntegrator.

Please advise if the solution in https://github.com/siconos/siconos/commit/3be9667fc2dc5c9975e13b6f28c2cad33933b692 is acceptable.

vacary:
It seems that your solution is correct. I think it is a side effect of our work on lazy initialization.

My question is: why was it working for a dynamical system that is created at the beginning of the simulation ?

vacary:
ok, I got it It has been written in the initialize of OSI that is calls only when we initialize the Simulation. 

@fperignon Most of the code that you have factorized there:
~~~
  DynamicalSystemsGraph::VIterator dsi, dsend;
  for(std11::tie(dsi, dsend) = _dynamicalSystemsGraph->vertices(); dsi != dsend; ++dsi)
  {
    if(!checkOSI(dsi)) continue;
    SP::DynamicalSystem ds = _dynamicalSystemsGraph->bundle(*dsi);
[...]

PR #121, opened 2017-04-03
[kernel] Move all DS insertion and initialization to NSDS::insertDyna…

radarsat1:
[kernel] Move all DS insertion and initialization to NSDS::insertDynamicalSystem()

This is an attempt to address issues raised in #119.  Any comments appreciated, it if it's ok to merge it relatively quickly it would resolve some initialization issues across OSIs in one go.

Basically, this proposes to move all DS working memory initialization to OSI::initializeDynamicalSystem, and call that from NonSmoothDynamicalSystem::insertDynamicalSystem, which then does all the work to insert the DS, set the OSI, and initialize the memory.  Common code across OSIs is shared in an internal function in OneStepIntegrator.

Comments requested for whether the NSDS is a good place to do this.  My argument is that Topology() just manipulates the graph and so should do little work, but the NSDS is responsible to manipulate the Topology, so maybe it is okay if it knows how to associate a DS and an OSI, and initialize a DS according to the OSI.  The other candidate to do this would be the Simulation.

### DIFF
--- a/control/src/Controller/CommonSMC.cpp
+++ b/control/src/Controller/CommonSMC.cpp
@@ -161,7 +161,7 @@ void CommonSMC::initialize(const Model& m)
   _SMC->nonSmoothDynamicalSystem()->setControlProperty(_interactionSMC, true);
   _SMC->nonSmoothDynamicalSystem()->topology()->setName(_interactionSMC, "Sgn_SMC");
   _simulationSMC->setName("linear sliding mode controller simulation");
-  _simulationSMC->insertIntegrator(_integratorSMC);
+  _simulationSMC->initializeOSIforDS(_integratorSMC, _DS_SMC, _SMC, t0);
   // OneStepNsProblem
   _OSNSPB_SMC->numericsSolverOptions()->dparam[0] = _precision;
   //    std::cout << _OSNSPB_SMC->numericsSolverOptions()->dparam[0] <<std::endl;
--- a/control/src/Controller/CommonSMC.cpp
+++ b/control/src/Controller/CommonSMC.cpp
@@ -155,8 +155,8 @@ void CommonSMC::initialize(const Model& m)
     _integratorSMC.reset(new ZeroOrderHoldOSI());
   }
 
-  _SMC->nonSmoothDynamicalSystem()->insertDynamicalSystem(_DS_SMC);
-  _SMC->nonSmoothDynamicalSystem()->topology()->setOSI(_DS_SMC, _integratorSMC);
+  _SMC->nonSmoothDynamicalSystem()->insertDynamicalSystem(_DS_SMC, _SMC,
+                                                          t0, _integratorSMC);
   _SMC->nonSmoothDynamicalSystem()->setName(_DS_SMC, "plant_SMC");
   _SMC->nonSmoothDynamicalSystem()->link(_interactionSMC, _DS_SMC);
   _SMC->nonSmoothDynamicalSystem()->setControlProperty(_interactionSMC, true);
--- a/control/src/Controller/CommonSMC.cpp
+++ b/control/src/Controller/CommonSMC.cpp
@@ -155,8 +155,7 @@ void CommonSMC::initialize(const Model& m)
     _integratorSMC.reset(new ZeroOrderHoldOSI());
   }
 
-  _SMC->nonSmoothDynamicalSystem()->insertDynamicalSystem(_DS_SMC, _SMC,
-                                                          t0, _integratorSMC);
+  _SMC->nonSmoothDynamicalSystem()->insertDynamicalSystem(_DS_SMC);
   _SMC->nonSmoothDynamicalSystem()->setName(_DS_SMC, "plant_SMC");
   _SMC->nonSmoothDynamicalSystem()->link(_interactionSMC, _DS_SMC);
   _SMC->nonSmoothDynamicalSystem()->setControlProperty(_interactionSMC, true);
--- a/control/src/Controller/LinearSMCOT2.cpp
+++ b/control/src/Controller/LinearSMCOT2.cpp
@@ -113,7 +113,7 @@ void LinearSMCOT2::initialize(const Model& m)
   _PhiOSI.reset(new LsodarOSI());
   _modelPhi->nonSmoothDynamicalSystem()->insertDynamicalSystem(_DSPhi);
   _simulPhi.reset(new EventDriven(_tdPhi, 0));
-  _simulPhi->insertIntegrator(_PhiOSI);
+  _simulPhi->initializeOSIforDS(_PhiOSI, _DSPhi, _modelPhi, _t0);
   _modelPhi->setSimulation(_simulPhi);
   _modelPhi->initialize();
   // Integration for Gamma
@@ -121,7 +121,7 @@ void LinearSMCOT2::initialize(const Model& m)
   _PredOSI.reset(new LsodarOSI());
   _modelPred->nonSmoothDynamicalSystem()->insertDynamicalSystem(_DSPred);
   _simulPred.reset(new EventDriven(_tdPred, 0));
-  _simulPred->insertIntegrator(_PredOSI);
+  _simulPred->initializeOSIforDS(_PredOSI, _DSPred, _modelPred, _t0);
   _modelPred->setSimulation(_simulPred);
   _modelPred->initialize();
 
--- a/control/src/Controller/LinearSMCOT2.cpp
+++ b/control/src/Controller/LinearSMCOT2.cpp
@@ -111,8 +111,7 @@ void LinearSMCOT2::initialize(const Model& m)
 
   _modelPhi.reset(new Model(_t0, _T));
   _PhiOSI.reset(new LsodarOSI());
-  _modelPhi->nonSmoothDynamicalSystem()->insertDynamicalSystem(_DSPhi, _modelPhi,
-                                                               _t0, _PhiOSI);
+  _modelPhi->nonSmoothDynamicalSystem()->insertDynamicalSystem(_DSPhi);
   _simulPhi.reset(new EventDriven(_tdPhi, 0));
   _simulPhi->insertIntegrator(_PhiOSI);
   _modelPhi->setSimulation(_simulPhi);
@@ -120,8 +119,7 @@ void LinearSMCOT2::initialize(const Model& m)
   // Integration for Gamma
   _modelPred.reset(new Model(_t0, _T));
   _PredOSI.reset(new LsodarOSI());
-  _modelPred->nonSmoothDynamicalSystem()->insertDynamicalSystem(_DSPred, _modelPred,
-                                                                _t0, _PredOSI);
+  _modelPred->nonSmoothDynamicalSystem()->insertDynamicalSystem(_DSPred);
   _simulPred.reset(new EventDriven(_tdPred, 0));
   _simulPred->insertIntegrator(_PredOSI);
   _modelPred->setSimulation(_simulPred);
--- a/control/src/Controller/LinearSMCOT2.cpp
+++ b/control/src/Controller/LinearSMCOT2.cpp
@@ -111,8 +111,8 @@ void LinearSMCOT2::initialize(const Model& m)
 
   _modelPhi.reset(new Model(_t0, _T));
   _PhiOSI.reset(new LsodarOSI());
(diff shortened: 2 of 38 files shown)


vacary:
Thank you for your work. 

I fully agree that the insertion process has to be improved. I have just one remark:

Is it possible to split the insertion into two parts to separate clearly modeling and simulation? The goal would be have an insertion into the non smooth dynamical system that is independent from the simulation and then to have an insertion in the simulation that deal with the osi and the initialization of the ds specific to the osi. By the way, the name that I chose for the method initializeDynamicalSystem is not sufficiently clear. I would prefer initializeForDynamicalSystem, because it is not a initialization of the Dynamical System but for the osi. The same remark applies to initializeInteraction.

The separation of simulation and modeling is of utmost importance in Siconos. I know that the Topology class is something that mixes both but if we can avoid at most at possible this mix, I think it is better.

What do your think ?

fperignon:
> Le 4 avr. 2017 à 08:55, Vincent Acary <notifications@github.com> a écrit :
> 
> Thank you for your work.
> 
> I fully agree that the insertion process has to be improved. I have just one remark:
> 
> Is it possible to split the insertion into two parts to separate clearly modeling and simulation? The goal would be have an insertion into the non smooth dynamical system that is independent from the simulation and then to have an insertion in the simulation that deal with the osi and the initialization of the ds specific to the osi. By the way, the name that I chose for the method initializeDynamicalSystem is not sufficiently clear. I would prefer initializeForDynamicalSystem, because it is not a initialization of the Dynamical System but for the osi. The same remark applies to initializeInteraction.
> 
> The separation of simulation and modeling is of utmost importance in Siconos. I know that the Topology class is something that mixes both but if we can avoid at most at possible this mix, I think it is better.
> 
> What do your think ?
> 

yes,  we should not call any initialization process related to simulation (osi or else) from nsds functions.
When a new DS is inserted on the fly, maybe the best is to explicitly call some setOSI function, to link the ds with an osi AND to initialize all the
work buffers (osi related). 
Topology is still messy and maybe we should think of splitting  it into « nsds-topology » with DS and Interaction index set 0 and « simu-topology »  dealing with index sets i, i >0.

radarsat1:
> yes, we should not call any initialization process related to simulation (osi or else) from nsds functions.

Fair enough.  I chose to try in the NSDS since it is shared between the simulation and the model.  During insertion there was already a call to Topology::setOSI() everywhere, and I thought maybe it was exposing the Topology too much in user code.  Perhaps this could be called by a function in the Simulation.

I suppose the fundamental difficulty is that inserting a DS affects both the Model and the Simulation.  Thus one would expect 2 calls needed to insert a DS, Simulation::insertDynamicalSystem and Model::insertDynamicalSystem.  However one does not want to insert the same DS twice into the shared NSDS, so some coordination is needed.

I would propose the following then:
* Keep NSDS::insertDynamicalSystem (no need to move it to Model)
* Replace calls to Topology::setOSI with Simulation::initializeOSIforDS(ds, osi)  (which calls setOSI as well as initializes work vectors -- unclear if DS::resetToInitialState should be called)

The only "hitch" is that OSI::initializeDynamicalSystem takes the model and time as argument, reinforcing a link from Simulation to Model.  However, I'll note that these two arguments are rarely used. Model is only used in ZeroOrderHoldOSI, although time is used in MoreauJeanOSI and derivatives, without reference to Model.

fperignon:
> Le 4 avr. 2017 à 15:27, Stephen Sinclair <notifications@github.com> a écrit :
> 
> yes, we should not call any initialization process related to simulation (osi or else) from nsds functions.
> 
> Fair enough. I chose to try in the NSDS since it is shared between the simulation and the model. During insertion there was already a call to Topology::setOSI() everywhere, and I thought maybe it was exposing the Topology too much in user code. Perhaps this could be called by a function in the Simulation.
> 
> I suppose the fundamental difficulty is that inserting a DS affects both the Model and the Simulation. Thus one would expect 2 calls needed to insert a DS, Simulation::insertDynamicalSystem and Model::insertDynamicalSystem. However one does not want to insert the same DS twice into the shared NSDS, so some coordination is needed.
> 
> I would propose the following then:
> 
> Keep NSDS::insertDynamicalSystem (no need to move it to Model) 
yes. And this function add a vertex inside the ds graph, that’s it.
> Replace calls to Topology::setOSI with Simulation::initializeOSIforDS(ds, osi) (which calls setOSI as well as initializes work vectors -- unclear if DS::resetToInitialState should be called)
yes, something equivalent to the nsds->link(inter, ds) but for the osi.

simu->link(osi, ds) :
- insert osi in simu (i.e. call simu.insertIntegrator(osi))
- set osi property in the ds graph for the inserted ds (i.e. former topology::setOSI)
- « prepare » the osi for the simulation (allocate/init buffers, reset and so on)

> The only "hitch" is that OSI::initializeDynamicalSystem takes the model and time as argument, reinforcing a link from Simulation to Model. However, I'll note that these two arguments are rarely used. Model is only used in ZeroOrderHoldOSI, although time is used in MoreauJeanOSI and derivatives, without reference to Model.
> 
> 
Let us try and fix thereafter ...

vacary:
+1

radarsat1:
I updated the branch.  I named the function initializeOSIforDS but I'm happy to rename it to link() if you prefer, however I think link() is better reserved for DS/Interaction linking as it operates on the graph, and it's good to indicate in the name that some memory is initialized.  We could change initializeOSIforDS to something like setOSIforDS or assignOSItoDS().. not sure..

Currently it does not call insertIntegrator().

control/src/Controller/CommonSMC.cpp
```diff
@@ -156,7 +156,6 @@ void CommonSMC::initialize(const Model& m)
   }
 
   _SMC->nonSmoothDynamicalSystem()->insertDynamicalSystem(_DS_SMC);
-  _SMC->nonSmoothDynamicalSystem()->topology()->setOSI(_DS_SMC, _integratorSMC);
```
xhub: Shouldn't this be replaced by a call to initializeOSIforDS?

xhub: Shouldn't this be replace by a call to initializeOSIforDS?

control/src/Controller/LinearSMCOT2.cpp
```diff
@@ -112,7 +112,6 @@ void LinearSMCOT2::initialize(const Model& m)
   _modelPhi.reset(new Model(_t0, _T));
   _PhiOSI.reset(new LsodarOSI());
   _modelPhi->nonSmoothDynamicalSystem()->insertDynamicalSystem(_DSPhi);
-  _modelPhi->nonSmoothDynamicalSystem()->topology()->setOSI(_DSPhi, _PhiOSI);
```
xhub: ditto

kernel/src/modelingTools/NonSmoothDynamicalSystem.cpp
```diff
@@ -19,6 +19,7 @@
 #include "Interaction.hpp"
 #include "LagrangianLinearTIDS.hpp"
 #include "FirstOrderLinearTIDS.hpp"
+#include "OneStepIntegrator.hpp"
```
xhub: Is this still necessary?

kernel/src/modelingTools/NonSmoothDynamicalSystem.hpp
```diff
@@ -111,11 +111,7 @@ class NonSmoothDynamicalSystem
   /** add a dynamical system into the DS graph (as a vertex)
    * \param ds a pointer to the system to add
    */
-  inline void insertDynamicalSystem(SP::DynamicalSystem ds)
-  {
-    _topology->insertDynamicalSystem(ds);
-    _mIsLinear = ((ds)->isLinear() && _mIsLinear);
-  };
+  void insertDynamicalSystem(SP::DynamicalSystem ds);
```
xhub: Is there a change to insertDynamicalSystem other than make it non-inline. I'm by far not against making this one non-inline, espescially when we can get rid of some include in the header files.

kernel/src/simulationTools/test/ZOHTest.cpp
```diff
@@ -40,7 +40,6 @@ void ZOHTest::init()
   _sim.reset(new TimeStepping(_TD, 0));
   _ZOH.reset(new ZeroOrderHoldOSI());
   _model->nonSmoothDynamicalSystem()->insertDynamicalSystem(_DS);
-  _model->nonSmoothDynamicalSystem()->topology()->setOSI(_DS, _ZOH);
```
xhub: shoudn't there be a call to initializeOSIforDS?

control/src/Controller/CommonSMC.cpp
radarsat1: Hm, in this case I skipped it because in fact it is too early for initializeOSIforDS to call initializeDynamicalSystem, which is okay to defer to Model/Simulation::initialize(), but I forgot that this would skip the OSI association. In fact Simulation::initialize() assigns the first OSI if it is not done prior.  It would be nice to do the same thing in *all cases* from the user code point of view, so perhaps it can call initializeOSIforDS() in any case, and it will simply skip initializeDynamicalSystem if the OSI is not yet initialized.  Then initializeDynamicalSystem will be called by OSI::initialize() during Simulation::initialize().

kernel/src/modelingTools/NonSmoothDynamicalSystem.cpp
radarsat1: Probably not, I'll revert.

kernel/src/modelingTools/NonSmoothDynamicalSystem.hpp
radarsat1: I'll just revert insertDynamicalSystem to how it was for now.  I don't think it will allow removing an include.

control/src/Controller/CommonSMC.cpp
radarsat1: Oh right, but then this imposes an ugly and useless need to create the Simulation object before adding DSs to the Model, which is not how many of the examples work.  I'll do it anyways for now just to see how it looks, but it seems like a wrong turn.  Otherwise we go back to calling topology()- setOSI() manually, but then have two distinct cases:

1. DS is created before the Simulation, call insertDynamicalSystem and nsds->topology()->setOSI
2. DS is created after the Simulation, call insertDynamicalSystem and Simulation::initializeOSIforDS.

Confusing.

radarsat1: (sorry, still thinking...) On the other hand if we followed @fperignon's idea to have it call insertIntegrator(), then it would make sense for initializeOSIforDS to be placed where insertIntegrator is placed, instead of where setOSI is placed.  This is nicer, as the Simulation and OSI must already exist at that point.

radarsat1:
Simulation::initializeOSIforDS now also replaces Simulation::insertIntegrator.  Comments welcome before I rebase this on master.  I think it will be more clear to rebase than to merge in this case.

vacary:
I think with we can merge now.

@radarsat1 Do you want to merge or rebase ?

radarsat1:
I think it would be more clear to rebase in this case.  I'll apply to master today if we're all happy with initializeOSIforDS().  (We can still change it after of course..)

kernel/src/simulationTools/Simulation.cpp
```diff
@@ -497,3 +497,19 @@ void Simulation::updateOutput(unsigned int)
...
+  insertIntegrator(osi);
+
+  // Associate the OSI to the DS in the topology.
+  _nsds->topology()->setOSI(ds, osi);
+
+  // If no Model, or OSI has no DSG yet, assume DS will be initialized
+  // later.  (Typically, during Simulation::initialize())
+  if (m && osi->dynamicalSystemsGraph())
+    osi->initializeDynamicalSystem(*m, time, ds);
+}
```
xhub: I think it would make sense to have default argument for m and time, so that it is (imho) cleaner. The default time could come from the model. This brings also a question: when does time different from the currentTime in the Model?

radarsat1: I'll add default m==null, is that what you mean?

xhub: Nope, it is a share_ptr, so it should be

std11::shared_ptr<Model>()

but there may be something simpler ... that's the best I could come up with.

radarsat1:
Closed in 88718c513426e06c023f27c3a505d67b1c758542

Function renamed to prepareIntegratorForDS

PR #139, opened 2017-05-11
Fix OCC.

radarsat1:
I have never been able to get OCC/mechanisms to compile, let alone get the tests running. So I finally decided to check it out.  These were the changes I had to make.  The only one I am not sure about is the call to math_BrentMinimum, which seems to be different from the API specified for either OCC [6.9.0](https://www.opencascade.com/doc/occt-6.9.0/refman/html/classmath___brent_minimum.html) or [7.1.0](https://www.opencascade.com/doc/occt-7.1.0/refman/html/classmath___brent_minimum.html).

For users of mechanisms, please check that this doesn't break something, maybe it was different on old versions.  The environment I have used for testing is from https://github.com/siconos/siconos-tutorials/blob/master/Dockerfile-with-mechanisms and therefore uses OCE and PythonOCC from github.

### DIFF
--- a/mechanics/src/mechanisms/CADMBTB/mymath_FunctionSetRoot.cxx
+++ b/mechanics/src/mechanisms/CADMBTB/mymath_FunctionSetRoot.cxx
@@ -202,7 +202,8 @@ static Standard_Boolean MinimizeDirection(const math_Vector&   P0,
   ax = -1; bx = 0;
   cx = (P2-P1).Norm()*invnorme;
   if (cx < 1.e-2) return Standard_False;
-  math_BrentMinimum Sol(F, ax, bx, cx, tol1d, 100, tol1d);
+  math_BrentMinimum Sol(tol1d, 100, tol1d);
+  Sol.Perform(F, ax, bx, cx);
   if(Sol.IsDone()) {
     tsol = Sol.Location();
     if (Sol.Minimum() < F1) {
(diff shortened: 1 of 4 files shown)


fperignon:
> Le 11 mai 2017 à 20:45, Stephen Sinclair <notifications@github.com> a écrit :
> 
> I have never been able to get OCC/mechanisms to compile, let alone get the tests running. So I finally decided to check it out. There were the changes I had to make. The only one I am not sure about is the call to math_BrentMinimum, which seems to be different from the API specified for either OCC 6.9.0 <https://www.opencascade.com/doc/occt-6.9.0/refman/html/classmath___brent_minimum.html> or 7.1.0 <https://www.opencascade.com/doc/occt-7.1.0/refman/html/classmath___brent_minimum.html>.
> 

just one more info:
last time I tried to install mechanisms,  I need to downgrade oce/pythonoccI to use it with siconos :


oce —> git checkout OCE-0.16.1
pythonocc —> git checkout 0.16.5

did you succeed with the last version?

radarsat1:
Ah that explains my previous failures.  Sorry if it was mentioned somewhere and I did not notice.

But yes, with this PR, I am able to compile against git master versions of both repositories.  However, I can only confirm it compiles and that OccTest.cpp runs correctly.  I am not familiar enough with Mechanisms to test it fully.

PR #155, opened 2017-06-20
siconos_vview (small) enhancements

bremond:
An attempt on the visualisation of big problems (some chute con rocas). These enhancements
only provides a few % gain on a python cProfile with --no-cf option. The main bottleneck remains
the contact forces visualisation and apparently this is not due to python but to the way we use vtk for this.

In those file you can compare the cumulative time for set_position (beware set_positionv = set_position in new version) and set_actors_visibility

set_position (just 1% better)
 3.975/40.246 = 0.0988 (ori)
 3.285/38.801 = 0.0847 (new)

set_actors_visibility (much better!)
 2.311 / 40.246 = 0.0574 (ori)
 0.161 / 38.801 = 0.00415 (new)

Note that there is some bad initializations that disappear as soon as time is increased, this should be easy to fix without modifying set_actors_visibility.

[rprofile_enhancements.txt](https://github.com/siconos/siconos/files/1087633/rprofile_enhancements.txt)
[rprofile_ori.txt](https://github.com/siconos/siconos/files/1087635/rprofile_ori.txt)


bremond:
Sorry the [occ] stuff has nothing to do here, I reverted the commit...

radarsat1:
I just saw that some of the changes I made to today conflict with this, very randomly I noticed that only the first actor of a set of contactors was being made visible.  I'll try to merge the changes.

radarsat1:
There are some unfixed merge conflicts in vview.py in this branch, e.g. `<<<<<<< HEAD`, in 9c16e7b8771649272bacc678b6a87e72a7fb5bb7

It seems the fix I made today is incompatible here, sorry for the conflict, but I realized that each object has a list of actors, not a single actor, so the vectorization change may not work, do you think it can be adapted?

Also, with the presence of the revert on the [occ] commit, perhaps it would be best to rebase this PR.

I noticed your question,

```
          # Collect positions from init data
          # but why ??
```

the reason why I made this change was to be able to visualize not-yet started simulations, where there is not yet any data in the "dynamic" table, but the positions are nonetheless available in the metadata.

Otherwise the only way to visualize the initial conditions of a simulation is to execute the first time step, which is not so convenient when developing a script.

PR #160, opened 2017-06-22
vview.py: some (small) enhancements

bremond:
mainly in those 2 functions:
     - set_dynamic_actors_visibility
     - set_position

     + split actors dictionay in dynamic_actors + static_actors

     Note: after looking again at some visualisations on chute con rocas
     there is no doubt that this is the multiple actors per object that
     does not scale (and finally this is obvious to me now).

### DIFF
--- a/io/swig/io/vview.py
+++ b/io/swig/io/vview.py
@@ -188,36 +188,6 @@ big_data_writer.SetInputConnection(big_data_source.GetOutputPort())
 contactors = dict()
 offsets = dict()
 
-def set_position(instance, q0, q1, q2, q3, q4, q5, q6):
-    if (numpy.any(numpy.isnan([q0, q1, q2, q3, q4, q5, q6]))
-            or numpy.any(numpy.isinf([q0, q1, q2, q3, q4, q5, q6]))):
-        print('Bad position for', instance, q0, q1, q2, q3, q4, q5, q6)
-        return
-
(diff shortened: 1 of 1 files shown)


bremond:
Sorry, I made a new to branch to rebase everything and then I deleted vview_fixes branch in order to push the the rebased vview. Apparently, this made the previous PR disappear... So I understand that I should have created a new PR and close the previous one to keep the history.

io/swig/io/vview.py
```diff
@@ -918,22 +890,36 @@ def init_cf_sources(mu):
         transforms[instance] = []
         offsets[instance] = []
         actors[instance] = []
```
radarsat1: Just need to remove this line and then it works for me. ("NameError: name 'actors' is not defined")

io/swig/io/vview.py
```diff
@@ -918,22 +890,36 @@ def init_cf_sources(mu):
...
+        times_of_birth[instance] = io.instances()[instance_name].\
+                                   attrs['time_of_birth']
+
+        if instance > 0:
```
radarsat1: Is it "instance > 0" or "instance >= 0" for dynamic objects?

bremond: Argh.. actors[instance] should not exist as actors is splitted in dynamic_actors and static_actors, I miss this one when I rebased.
And yes, it is instance >= 0 for dynamic objects...

Issue #161, opened 2017-06-23
Poll: Should default joints be absolute or relative to q1?


radarsat1:
I am currently normalizing the API for joints, because there are still some outstanding differences in the constructors etc that make them confusing to use.  I'd like to make things the same and as least surprising as possible.

One difference is that some joint constructors have absoluteRef = false, some have absoluteRef = true, and some don't have absoluteRef and assume false.

I am adding the absoluteRef option to all joint constructors.  Should the default be absolute or relative to q1?

My vote is for absolute (i.e. absoluteRef=true)

@bremond @vacary

radarsat1:
Equally valid could be the opinion that absoluteRef should not have a default value!  Perhaps it would make sense to force the user to specify...

radarsat1:
After some thought, I think that last idea is best.  It should never be "automatic" whether P and A are relative to q1 or not, the user must specify.  I've coded this along with some other changes to joint construction in #162.

radarsat1:
I'll go ahead and merge this tomorrow if there is no argument.  I posed it as a PR just to make sure it does not surprise anyone or break code that is developed outside the Siconos repository.

But if you think such a breaking change should wait until 4.2 please let me know.

bremond:
Is it ok with mechanisms /MBTB stuff (https://github.com/siconos/siconos/tree/master/mechanics/src/mechanisms) ? Sorry I do not have to check right now (the SliderCrank for example)
No tests are there... and we can think of of an interface occ/new mechanics API for 4.2 but not for 4.1

radarsat1:
I checked the OCC examples, but I will double-check on mechanisms before merging anything.

radarsat1:
Yes, a few lines in mechanisms were affected, thanks.

radarsat1:
Closed in 77ecc5e4225d2d29242961975e5670b4550b6bd8

PR #162, opened 2017-06-27
joints construction: points and axes in superclass, allow deferred init

radarsat1:
I want to clean up the constructors for joints before the next release.  The main difficult point as mentioned in #161 is that some joints (Knee, Pivot..) default to absoluteRef=true while others (Prismatic) default to absoluteRef=false.  There are some other subtle differences, e.g. some joints initialize on `d1->q()` while others initialize on `d1->q0()`.  Also it is not always clear if the pointer to the provided vector will be retained, or if the data will be copied.

Thinking on this, I propose a few changes:
1. Separate joint "properties" (axis, points, absolute/relative..) from joint "initial conditions", i.e. position of DS.
2. Always force specifying absoluteRef, with no default.
3. However, allow joints to be created without specifying properties or initial conditions, or just by specifying properties and leaving initial conditions for later.
4. Retain the pointers to the provided SiconosVector, but only use it when the initial conditions are provided.  (i.e., not check for changes at every timestep..)

This patch makes the following changes:
1. Move handling of points and axes to two VectorOfVector objects in NewtonEulerJointR.
2. These VectorOfVectors are sized by the subclasses according to how many points and axes are needed.
3. A common interface setPoint() and setAxis() are provided in NewtonEulerJointR.
4. Calculation of initial conditions is done in a common function setInitialConditions(), which only takes SiconosVectors of size 7, no need to provide a NewtonEulerDS.  No longer have different constructors for the 1- and 2-DS conditions.
5. All constructors are reduced to calling setPoint/setAxis and setInitialConditions().
6. A default do-nothing constructor is publically available, the caller can call these functions him/herself.  The axis/point properties can be changed at any time but only take effect if setInitialConditions is called.
7. Order of arguments to the constructors is changed so that the 1-DS condition is just indicated by a default null pointer for second DS.

Some examples.

Construct a PivotJointR between a body and the absolute frame step by step:
~~~
SP::NewtonEulerDS body1, body2; // assume already exist..
SP::SiconosVector P(new SiconosVector(3));
SP::SiconosVector A(new SiconosVector(3));
PivotJointR joint();
joint.setPoint(0, P);
joint.setAxis(0, A);
joint.setAbsolute(true); // (this is default)
joint.setInitialConditions(body1->q());
~~~

Construct a PivotJointR between two bodies with a frame relative to body1:
~~~
PivotJointR joint(P, A, false);
joint.setInitialConditions(body1->q(), body2->q());
~~~

Construct a PivotJointR between two bodies with a frame relative to body1:
~~~
PivotJointR joint(P, A, false, body1, body2);
~~~

This patch contains changes to all JointsTests* examples to adapt the constructor call.

### DIFF
--- a/examples/Mechanics/JointsTests/NE_1DS_1Knee_MLCP.cpp
+++ b/examples/Mechanics/JointsTests/NE_1DS_1Knee_MLCP.cpp
@@ -166,7 +166,7 @@ int main(int argc, char* argv[])
     // Building the first knee joint for beam1
     // input  - the concerned DS : beam1
     //        - a point in the spatial frame (absolute frame) where the knee is defined P
-    SP::KneeJointR relation1(new KneeJointR(beam1, P));
+    SP::KneeJointR relation1(new KneeJointR(P, true, beam1));
     SP::NonSmoothLaw nslaw1(new EqualityConditionNSL(relation1->numberOfConstraints()));
 
     SP::Interaction inter1(new Interaction(nslaw1, relation1));
@@ -294,7 +294,7 @@ int main(int argc, char* argv[])
     std::cout << "Error w.r.t. reference file : " << (dataPlot - dataPlotRef).normInf() << std::endl;
     if ((dataPlot - dataPlotRef).normInf() > 1e-7)
     {
-      (dataPlot - dataPlotRef).display();
+      //(dataPlot - dataPlotRef).display();
       std::cout << "Warning. The results is rather different from the reference file." << std::endl;
       return 1;
     }
--- a/examples/Mechanics/JointsTests/NE_1DS_1Knee_MLCP_MoreauJeanCombinedProjection.cpp
+++ b/examples/Mechanics/JointsTests/NE_1DS_1Knee_MLCP_MoreauJeanCombinedProjection.cpp
@@ -163,7 +163,7 @@ int main(int argc, char* argv[])
     // Building the first knee joint for beam1
     // input  - the concerned DS : beam1
     //        - a point in the spatial frame (absolute frame) where the knee is defined P
-    SP::KneeJointR relation1(new KneeJointR(beam1, P));
+    SP::KneeJointR relation1(new KneeJointR(P, true, beam1));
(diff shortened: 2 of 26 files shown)


radarsat1:
I realize that there are a couple of points at odds in my description above:

> Always force specifying absoluteRef, with no default.

vs. 

> allow joints to be created without specifying properties

implies that there must be a default value for absoluteRef in the second case.  Currently it is true, as mentioned in a comment in the example.  However I think this is very useful, any ideas welcome..

Also, I left the constructor accepting NewtonEulerDS in order to match the previous API, which means that in the default call to setInitialConditions(), I do call it with ds1->q() and ds2->q().  If it is called by the user he can specify q0() instead, however I figured having a default case could be useful, as long as it is consistent across joints.

vacary:
I agree with all your changes. Just before to accept the pull request, I have a question

Could you explain what are the meaning of the initial conditions ? it seems that I missed something.

The joints are usually defined by constraints of the type h(Q) = 0 where Q concatenates the position of the concerned DSs. The vector should be provided by the "link" method

Why do we need to precise "initial conditions" ? is it for the possibility to define to define a joint with a static object ? Perhaps, it is only the term "initial conditions that is misleading ?

radarsat1:
Hm, perhaps initial conditions is not the right terminology, but basically apart from the points and axes that are used to define the joint, there are some constants to initialize based on the initial q() vectors of the DS.  These keep the original relative positions and orientation between the two bodies for example.

Previously this was done in the constructors but I decided to move it out to a function.  In fact now that it is moved out it could be called by link(), which wasn't possible before.

In that case the main problem is just an inheritance issue: link() takes the Interaction, which doesn't know that its relation is a NewtonEulerJointR.  It would need a specialization for the NewtonEulerJointR case, or we would need to have a virtual function in Relation that gets called during link() for all Interactions with pointers to the bodies, or at least the bodies q().

In any case, if not setInitialConditions(), what name would you give to the function that takes the initial q() vectors and calculates the constant position/orientation offsets?  updateOffsets()?  setInitialOffsets()?

There is the initialize_ds_links() route, which maybe could be used for this, accessing the initial q() from q0 in the DSLink. Is that preferable?

Issue #142, opened 2017-05-16
How to set an OSI for a given DS ?


vacary:
Following the comment of @radarsat1  on my commit f57b5ac9caf9fe0f5494f376fcb45841c30a5edc 
I think it could be useful to discuss how to set the OSI for a given DS

radarsat1:
Yes, I started adding calls to Simulation::setNonSmoothDynamicalSystemPtr() after the Simulation was created, because I figured that it didn't really make sense to have a Simulation without an NSDS anyways.  However, maybe it's a non-issue since prepareIntegratorForDS actually takes the Model as a parameter anyways, so we could just take the NSDS from there.  I didn't do it that way since the idea was that supplying the Model should be optional.  In principle we should also assert() that the Simultion::_nsds == Model::_nsnds.

In any case, the steps needed for adding a DS to the simulation and model are, as I understand them,

1. Add the DS to the graph using NSDS::insertDynamicalSystem()
2. Ensure the OSI is part of Simulation::_allOSI using Simulation::insertIntegrator()
3. Set the OSI for the DS via topology()->setOSI()
4. initialize DS work vectors using OSI::initializeDynamicalSystem(), which requires the model and time.

The ambiguity in the API comes from the fact that the NSDS is shared between Simulation and Model.  I put prepareIntegratorForDS() into the Simulation since it requires to access _allOSI, but since it also needs the NSDS which it not guaranteed for Simulation, we have to choose between the following options:

1. Take the NSDS from the Model, require the Model is provided.
2. Ensure Simulation has an NSDS
3. Move prepareIntegratorForDS to Model, provide Simulation as argument.

I have been doing (2), but we could do (1) instead.  (3) does not make sense to me.

vacary:
I am ok with the 4 steps you described for adding a DS into the simulation.

For the API, perhaps we can revert one step by implementing a method such as

~~~
OneStepIntegrator::addDynamicalSystemToIntegrate(DynamicalSystem & ds)
{
topology()->setOSI( *this, ds);
}
~~~

If the first two steps has been done successfully, then the topology is reachable but the OSI.

@radarsat1 What do you think ?

radarsat1:
Hm, I didn't realize the OSI also has a pointer to the NSDS.  Could be interesting, I will think a bit more.  My first impression is that it doesn't sound very intuitive to "add" the DS to the OSI, but in fact if the OSI needs workspace for each DS then maybe that makes sense.

vacary:
[...]

PR #166, opened 2017-07-06
[kernel] make prepareIntegratorForDS require the Model.

radarsat1:
Makes Simulation::setNonSmoothDynamicalSystemPtr() and explicit calls
to setOSI() unnecessary.

Attempt to stabilise the API for DS/OSI link initialiation, per issue #142, comments welcome.

### DIFF
--- a/kernel/src/simulationTools/Simulation.cpp
+++ b/kernel/src/simulationTools/Simulation.cpp
@@ -544,16 +544,37 @@ void Simulation::prepareIntegratorForDS(SP::OneStepIntegrator osi,
                                         SP::DynamicalSystem ds,
                                         SP::Model m, double time)
 {
+  assert(m && m->nonSmoothDynamicalSystem() && "Simulation::prepareIntegratorForDS requires a Model with an NSDS.");
+
+  /*
+   * Steps to be accomplished when adding a DS to a Model and
+   * Simulation:
+   *
+   * 1. Add the DS to model->_nsds (Model::insertDynamicalSystem(ds))
+   *    (assumed done before this function is called, everything else
+   *    done in this function)
+   *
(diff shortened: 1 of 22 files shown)


kernel/src/simulationTools/Simulation.hpp
```diff
@@ -405,7 +405,7 @@ class Simulation : public std11::enable_shared_from_this<Simulation>
    *  \param m The Model for initializing the OSI.
    *  \param time The current time for initializing the OSI. */
   void prepareIntegratorForDS(SP::OneStepIntegrator osi, SP::DynamicalSystem ds,
-                              SP::Model m=SP::Model(), double time=0);
+                              SP::Model m, double time);
```
xhub: nitpick: we could have Model&m instead of SP::Model.
The use of a SP is unnecessary here. With a reference we also don't have to check whether it is non-empy

kernel/src/simulationTools/Simulation.cpp
```diff
@@ -544,16 +544,37 @@ void Simulation::prepareIntegratorForDS(SP::OneStepIntegrator osi,
                                         SP::DynamicalSystem ds,
                                         SP::Model m, double time)
 {
+  assert(m && m->nonSmoothDynamicalSystem() && "Simulation::prepareIntegratorForDS requires a Model with an NSDS.");
+
```
xhub: The check m->nonSmoothDynamicalSystem() should result in a exception, since m->nonSmoothDynamicalSystem()->topology()->setOSI(ds, osi); is going to fail otw. I'd do
if (!m->nonSmoothDynamicalSystem()) { RAISE EXCEPTION }

xhub:
I like this PR, it makes things a bit cleaner. I've just 2 suggestions

PR #171, opened 2017-07-07
[kernel] move Simulation::prepareIntegratorForDS to NSDS::setOSI taking Model param

radarsat1:
This is an alternative suggestion to PR #166, a different proposal for #142.  It's a lot more similar to what we had before, except that NSDS::setOSI calls both Topology::setOSI and OSI::initializeDynamicalSystem(), instead of doing it in the SImulation.  This means NSDS::setOSI requires the Model, similar to #166, but it means the DS/OSI link can be set up without creating the Simulation prior, as originally.  It's exactly like calling Topology::setOSI, but with the additional guarantee that OSI work vectors will be initialized.

As mentioned in #142, ZeroOrderHoldOSI::initializeDynamicalSystem requires the Model, an example of why it can't be removed as a parameter, so here instead it is added as a parameter to setOSI.  I also add the OSI as an optional parameter to NSDS::insertDynamicalSystem, so that a DS can be inserted and assigned an OSI in one line, which is the typical use case for multi-OSI examples.

### DIFF
--- a/control/src/Controller/CommonSMC.cpp
+++ b/control/src/Controller/CommonSMC.cpp
@@ -83,7 +83,6 @@ void CommonSMC::initialize(const Model& m)
   _SMC.reset(new Model(t0, T));
   // Set up the simulation
   _simulationSMC.reset(new TimeStepping(_td));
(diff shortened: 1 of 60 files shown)


radarsat1:
Closed without merge in favour of #166

PR #183, opened 2017-08-30
Openblas issue fix

bremond:
if openblas library is found, cblas.h is not searched anymore, it is the one provided under externals/blas_lapack/cblas_header

for more information see @xhub d28d3ca1be273094bb3eb06a826b357c71c01afd

the only modification to this commit is to always use the provided (and official) cblas.h

The compilation is ok on my ubuntu 17.04 + openblas:

-- Try to find blas in openblas ...
-- Looking for cblas_sgemm
-- Looking for cblas_sgemm - found
-- Looking for Fortran sgemm
-- Looking for Fortran sgemm - found
-- Blas libraries have been found : /usr/lib/libopenblas.so. We now turn to headers.
-- pkghints:
-- BLAS_INC_DIR:/usr/include
-- Found a library with BLAS API (openblas).
    Blas libraries : /usr/lib/libopenblas.so.
    Blas headers : cblas.h in /home/maurice/src/git/siconos/externals/blas_lapack.

### DIFF
--- a/cmake/FindBLAS.cmake
+++ b/cmake/FindBLAS.cmake
@@ -249,7 +249,7 @@ if(NOT BLAS_FOUND)
       # Right now openblas ships with a cblas.h header which uses double* as type for pointer to double complex, which makes ublas cry since it
       # want to pass std::complex<double>*
       set(BLAS_INCLUDE_SUFFIXES cblas_header)
-      set(INCLUDE_DIR_HINTS ${CMAKE_SOURCE_DIR}/externals/blas_lapack)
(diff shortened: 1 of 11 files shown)


radarsat1:
This works for me.

xhub:
To use the file in externals/blas_lapack/cblas_header is a hack, but it should work. I wrote the initial fix, so I'd rather continue use it (not surprisingly). However, if there is a better and more robust solution, it would be better.

xhub:
Is this still a current PR?

bremond:
2 pr were made at the same time fo this problem and #181 is the one that was merged.

PR #186, opened 2017-11-09
[kernel,examples] Put ref file comparison code in ioMatrix.

radarsat1:
I was thinking about packaging examples, and it seems to me that the ref files are used to double the examples as tests, which is okay, but for distribution they take a lot of room.

```
# cd ~/projects/debian/siconos/examples
$ du -sh .
268M	.

# remove ref files
$ find . -name \*.ref -delete

$ du -sh .
112M	.

$ find . -type f ! -name \*.cpp -a ! -name \*.c -a ! -name \*.py  -a ! -name \*.hpp -a ! -name \*.h -delete

$ du -sh .
5,5M
```

Not sure about removing everything else, but I propose that examples should run if the ref files are not present.  This PR proposes standardizing the ref-file comparison code and moving it into ioMatrix. I haven't done the work to change all examples to use this function yet, but I would like feedback on the idea.

It adds the function ioMatrix::compareRefFile, and one example of usage in DiodeBridgeCapFilter (chosen at random), changing the end of the program to,

```C++
    // --- elapsed time computing ---                                                    
    cout << "time = " << t.elapsed() << endl;                                            
                                                                                         
    // Number of time iterations                                                         
    cout << "Number of iterations done: " << k << endl;                                  
                                                                                         
    // dataPlot (ascii) output                                                           
    ioMatrix::write("DiodeBridgeCapFilter.dat", "ascii", dataPlot, "noDim");             
                                                                                         
    // Comparison with a reference file                                                  
    double error=0.0, eps=1e-12;                                                         
    if (ioMatrix::compareRefFile(dataPlot, "DiodeBridgeCapFilter.ref", eps, error)       
        && error > eps)                                                                  
      return 1;
```

In other words, if the ref file is available, then the program exits with 0 or 1 depending on the comparison.  Otherwise if the ref file is not available, the program exits normally and prints a warning.

### DIFF
--- a/examples/Electronics/DiodeBridgeCapFilter/DiodeBridgeCapFilter.cpp
+++ b/examples/Electronics/DiodeBridgeCapFilter/DiodeBridgeCapFilter.cpp
@@ -235,32 +235,10 @@ int main(int argc, char* argv[])
     ioMatrix::write("DiodeBridgeCapFilter.dat", "ascii", dataPlot, "noDim");
 
     // Comparison with a reference file
-    std::cout << "Comparison with a reference file" << std::endl;
-    SimpleMatrix dataPlotRef(dataPlot);
-    dataPlotRef.zero();
-    ioMatrix::read("DiodeBridgeCapFilter.ref", "ascii", dataPlotRef);
-    SP::SiconosVector err(new SiconosVector(dataPlot.size(1)));
-    (dataPlot - dataPlotRef).normInfByColumn(err);
-    err->display();
-    double error = 0.0;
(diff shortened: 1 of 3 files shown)


xhub:
It would be nice to pass an Index as an optional argument. Sometimes, some quantities are very volatile (like multipliers) and it doesn't make much sense to consider them in the comparison. In Control/Relay/Twisting.cpp, the comparison does not take into account the multipliers, but rather time and the state variables. Otherwise, that's a good idea.

radarsat1:
Added the index argument in 7fcfdfab0e7347c90e034b8b3f9de4b7aade9aa8

PR #188, opened 2017-11-17
Make optional installation of headers from "externals"

radarsat1:
This is related to packaging and might take some explanation.  The problem: Siconos has the "externals" library, which includes a lot of code used to make the solvers work, etc.  This is okay.  However, from a packaging point of view it strikes me as weird that Siconos re-exports much of these libraries as API, headers, and symbols.  For example, /usr/include/siconos has a bunch of stuff like f2h.h that are used internally to compile and link against Fortran libraries -- user code should not ever need to know about or use it.

So these two patches remove from installation targets firstly every header in externals other than CSparse, and then, secondly, CSparse-related headers.  I added options for both of these.

I made an exception for CSparse because it is different: the data structures and types defined in csparse.h are actually used in function parameters for Siconos functions, such as in NumericsMatrix.h and NumericsSparseMatrix.h.  As such, CSparse actually is part of the Siconos API.  So I was not sure whether or not to remove it.

The normal thing to do in such a case is to allow the user to define the functions without having the data structures, and he can include "csparse.h" himself if he wants to use them.  This is done by replacing the full struct definition with a prototype "struct cs_sparse", which is only access used through an "opaque" pointer.  So Siconos remains usable even if csparse.h is not installed, although for sparse matrices the user must have csparse.h.  (Typically, this should be a separately packaged library.)

However, since we are using an internal csparse, I did not dare take it out by default, so there are options to enable/disable this.  I was wondering if anyone has opinions on this before pushing it.  I have tested that all tests and examples still compile and run.

Anyways, the idea is that the libraries in externals are *used* by Siconos, but are not re-exported for Siconos users to access directly.  Thoughts?  Thumbs up/down?

### DIFF
--- a/cmake/default_options.cmake
+++ b/cmake/default_options.cmake
@@ -53,6 +53,9 @@ ELSE(UNIX)
   set(siconos_python_install "standard" CACHE STRING "Install mode for siconos python package")
 ENDIF(UNIX)
 
+# If OFF, headers from libraries in externals will not be installed.
+option(INSTALL_EXTERNALS_HEADERS
+  "Whether or not headers for external libraries should be installed. Default=ON" ON)
 
 # List of components to build and installed
 # List of siconos component to be installed
--- a/externals/CMakeLists.txt
+++ b/externals/CMakeLists.txt
@@ -57,6 +57,11 @@ if(NOT INSTALL_EXTERNALS_HEADERS)
     sort
     tools)
 endif()
+if(NOT INSTALL_EXTERNALS_HEADERS_CSPARSE)
+  list(APPEND ${COMPONENT}_HDRS_EXCLUDE_DIR
+    SuiteSparse/CSparse
+    SuiteSparse/CSparse_extensions)
+endif()
 
 # List of directories for which no doxygen doc will be generated
 # By default all directories matching "test" are excluded.
--- a/externals/CMakeLists.txt
+++ b/externals/CMakeLists.txt
@@ -46,6 +46,17 @@ set(${COMPONENT}_Unstable_SRCS)
 
 # List of directories of headers not to be installed
 set(${COMPONENT}_HDRS_EXCLUDE_DIR PATH_SDK/include SOL/lumod-c)
+if(NOT INSTALL_EXTERNALS_HEADERS)
+  list(APPEND ${COMPONENT}_HDRS_EXCLUDE_DIR
+    blas_lapack
+    hairer
+    netlib/dftemplates
+    netlib/odepack
+    optim_misc
+    optim_misc/ql0001
+    sort
+    tools)
+endif()
 
 # List of directories for which no doxygen doc will be generated
 # By default all directories matching "test" are excluded.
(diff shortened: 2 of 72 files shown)


radarsat1:
*TLDR*:

Two issues: 
1. do users of Siconos *need* access to `hairer.h` and things like that?
2. do users of Siconos *need* access to csparse.h?

(If 2 = yes, it's a strong argument for upgrading to the version of CSparse available in Debian, so maybe that's worth looking into.)

vacary:
Thank you very much. Very good job.

The users do not need to access to hairer.h. We can go further since haired code has no licence is to propose to compile with it only if we found it with a CPP key.

For csparse, it is not so clear. We can imagine that the user creates csparse matrix to fill some NumericsMatrix structure. We can also imagine that we rely on the suiteSparase distribution and we use it externally. I do not if we need a customized version of csparse but I don't think so.

vacary:
I just have  a look on the file cs.h distributed by suitesparse. It seems that the management of the type of entries in matrix has been completely change to render it more explicit.

If we want to update to this version it may be a huge but important work.

radarsat1:
Yes, that was my estimate as well, that updating would not be completely trivial.  I don't use numerics by itself very much, however after a glance in NumericsMatrix.h, it seems like the CSparse-related functions e.g. `NM_csc`, are one way to allocate/export sparse matrices, the other way is more "hidden" behind the NumericsMatrix interface, i.e., `NM_new` + `NM_csc_alloc` for example.  So if the latter is used more (externally), and the CSparse interface can be considered "optional", i.e. one can use NumericsMatrix just fine without the full definition of `cs_sparse`, then it seems okay to expose only opaque pointers to it.  In that case it is up to the user to use the same version of csparse.h, which we provide in the source distribution, but not in the libsiconos-numerics-dev package.  Then it is not an emergency to upgrade to the latest version, we can do it for 4.2 or 4.3.

If on the other hand csparse.h is "not optional" for using NumericsMatrix correctly, then we simply include it in the libsiconos-numerics-dev package in `/usr/include/siconos/csparse.h` and make the argument that this is the version we are currently using and it is necessary for now.  In any case it is modified from the original SuiteSparse so perhaps there is no problem.  We might take the effort to re-name things e.g. really search and replace `struct cs_sparse` for `CSparseMatrix`, or order to make it more "part of Siconos."  But it would be preferable to just upgrade and put a dependency on the libsuitesparse-dev package.

> If we update, we will be able at the same time to use the collections of linear solvers distributed with suitesparse.

Very good point!

vacary:
Should we put the upgrade to Suitesparse in a new issue ?

radarsat1:
Yes.  My impression from this exchange is that the first patch is useful, but removing suitesparse is too soon, it would be better to upgrade first.  I'll merge the first patch and close this, and open an issue.  I think I can find time next week to take a look at suitesparse.

radarsat1:
Closed in a591d40f31f2931a3be425db0a7cf1968860da07

radarsat1:
Related issue created in #189

PR #193, opened 2017-12-11
mechanics-occ

bremond:
It is an update for occ bindings.

Although this is not finished. It is usable now and I think we should sync before there are too many differences in mechanics_io and vview.

This provides:

 - occ_slidercrank.py with plugins and boundary conditions.
 - an inertia matrix computation from open cascade (compute_inertia_and_center_of_mass). A Volume class may be used to define volumes with displacements associated with material densities.
 - add all interactions between all contactors of two shapes, if contactors not precised.

some important renaming in mechanics_io & vview to avoid confusions : contactor.name becomes contactor.shape_name (with occ it becomes important to define multiples contactors from the same shape). Note that a compatibility function is provided : upgrade_io_format(filename) to modify old hdf5 files.

### DIFF
--- a/examples/Mechanics/OCC_Examples/occ_bouncing_ball.py
+++ b/examples/Mechanics/OCC_Examples/occ_bouncing_ball.py
@@ -4,7 +4,7 @@
 # Example of a bouncing ball with OpenCascade contactors
 #
 
-from siconos.mechanics.collision.tools import Contactor
+from siconos.mechanics.collision.tools import Volume, Contactor
 from siconos.io.mechanics_io import Hdf5
 from siconos import numerics
 import siconos.io.mechanics_io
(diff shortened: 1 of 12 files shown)


radarsat1:
I just had to fix one bug, the value of birth was wrong when calling importObject() which caused the value of _model.simulation() to be wrong since setSimulation() is not called yet.  (Swig doesn't make this None when the pointer is zero which made me take a lot longer to understand what was happening..)

After that, I am able to run simulations as before, so I pushed the merge.

bremond:
Ok thank you. I made some modifications for the occ bindings to be
compatible with time of birth, but I have not tested it (neither bullet
nor occ). I did nothing for time of death so it should fail with occ.

One another thing I didn't say is that occ simulation now works with a
cold restart (it was not possible before).

An important thing to remember is that these bindings, for the moment,
need the original api and it must be moved to the proposed
one. Probably, this sould need an issue in order not to forget it. Do you think
it is also time to remove completely the original api ?

radarsat1:
> Do you think it is also time to remove completely the original api ?

I think if it doesn't yet fit comfortably under the idea of SiconosShape etc., then maybe it is too early.  However, since those ideas in Mechanics (BodyDS, SiconosContactor, SiconosShape..) tend to fit well with MechanicsIO (contactor, object...), in principle it should be possible.  In general the idea with the "proposed API" was to more or less model the "mechanics_io" API on the C++ side, so thinking along those lines maybe we can come up with a way to fit the OCC work into that framework.  I might be easier when we get a chance to work in person, but in the meantime I can try to look more closely at what you've done so far and the OCC API to get an idea of how to fit things together better.

PR #194, opened 2017-12-18
[Numerics] try to fix SICONOS_INT64 problem

vacary:
This PR is not really a fix but a starting point for a discussion about the choice of integers for siconos sparse matrix.

On MacOsX 64 bit, the int64_t type is equivalent to a `long long int` integer. Although the size of `long long int` is the same of `long int` (8 bytes), we cannot perform such as allocation
`long * foo = (int64_t *) malloc(sizeof(int64_t));`
but 
`long long * foo = (int64_t *) malloc(sizeof(int64_t));` is valid

Since `CS_INT` is defined as a `long`, the allocation performed in Numerics_NM_Python.i is not compiling. In the proposed commit, I use CS_INT directly but it is not really satisfactory since we want to check if NPY_INT32 or NPY_INT64 corresponds to CS_INT.

One of my question is : Why we are using  `long` instead of `int64_t` if `SICONOS_INT64` is defined ?


radarsat1:
I can answer the last question at least,

> Why we are long instead of int64_t if SICONOS_INT64 is defined ?

Because that's how it is defined in cs.h.  Previously CSparse was modified to use int64_t.  If we want to avoid modifying it (which I believe we do), we have to adapt to it... that means using CS_INT whenever we are using a CSparse struct.  In other words, I think your change is probably correct.

On the other hand i am surprised it gives you an error, as I have been testing with both gcc and clang on Linux and Mac, and I only get a warning for this.  I did see the warning though, it was something I wasn't sure about, seeing as `long long` is 8 bytes as you say.  We have to make sure the conversion between numpy and CXSparse is correct, so good to bring this up.

vacary:
@radarsat1 Thank you for your answer.

Let us be compliant by cs.h but how can we be sure that the NPY_INT32 and NPY_INT64 will be correctly align with CS_INT. We need to change this code to be sure that it will work.

Documentation of numpy says:

NPY_INT32
The enumeration value for a 32-bit/4-byte signed integer.

NPY_LONG
Equivalent to either NPY_INT or NPY_LONGLONG, depending on the platform.

NPY_LONGLONG
NPY_INT64
The enumeration value for a 64-bit/8-byte signed integer.

So if we put CS_INT it should work on any platform with the `# ifdef _SICONOS_INT64`

mechanics/swig/mechanics/occ.i
```diff
@@ -126,51 +125,48 @@ typedef Interaction Interaction;
   }
 %}
 
+<<<<<<< Updated upstream
+=======
 %feature("nodirector") Geometer;
```
xhub: looks like a rebase artifact

mechanics/swig/mechanics/occ.i
```diff
@@ -126,51 +125,48 @@ typedef Interaction Interaction;
...
-PY_FULL_REGISTER(Geometer);
+PY_FULL_REGISTER(Geometer, Mechanics);
 
+>>>>>>> Stashed changes
 %feature("nodirector") ContactShapeDistance;
```
xhub: ditto

xhub (approved):
It seems sensible to do that. Just take care of the conflicts artifacts.

Now a long int is only guaranteed to be 32 bits long ... but we may worry about that another day.

PR #195, opened 2017-12-29
[numerics] refactoring of NumericsSparseMatrix and CSparsematrix

vacary:
add new method for reading sparse matrices from files


xhub (approved):
Looks good, but fails to build:

```
> CMake Error at /docker-workdir/numerics/src/tools/test/CMakeLists.txt:84 (ADD_EXECUTABLE):
>   Cannot find source file:
> 
>     /home/travis/build/siconos/siconos/numerics/src/tools/test/NumericsMatrix_convert.c
```
and
```

/.../numerics/src/tools/NumericsSparseMatrix.c:28:30: fatal error: Numerics_verbose.h: No such file or directory


/.../numerics/src/tools/CSparseMatrix.c:256:13: error: implicit declaration of function 'strtok_r' [-Werror=implicit-function-declaration]
```

radarsat1:
Happy new year!  Just back to work and testing this.  So far so good, test suite passes on my machine (that is, doesn't introduce new problems) so I'll push it.  I do agree it makes sense to move this to numerics.

Issue #22, opened 2016-03-12
[cmake] add a WITH_FORTRAN option


xhub:
motivations:
- not all platform have a fortran compiler easily available (Windows, Mac OS X)
- not all build system support fortran (Ninja, VS, ...)

We would loose at least:
- ql0001
- odepack
- hem5

Todo:
- add a cmake option WITH_FORTRAN
- and a bunch of "if defined(WITH_FORTRAN) ... else ... endif" in
  - LsodarOSI.{h,c}pp
  - Hem5OSI.{h,c}pp
  - lcp_nsqp.c, lcp_qp.c

Optional:
- add sundials (CVODE) to have a good event-driven integrator even if WITH_FORTRAN=NO

xhub:
Unfortunately, I don't have time to implement this now. This is low priority and should not block a release.

vacary:
fixed in #199

PR #199, opened 2018-01-16
[CMakeList.txt] add the option to compile with or without Fortran

vacary:
This PR tries to answer to the issue #22


radarsat1:
This works well for me.  Also, allows to build with `-g Ninja`!  (Actually I had to make one small change that I will submit.)  Any reason to be wary of merging it?  I don't see anything dangerous.  @xhub?

I would add to the documentation a clear list of things (solvers..) that will not work with the flag disabled.

CMakeLists.txt
```diff
@@ -39,6 +39,7 @@ if(USER_OPTIONS_FILE)
   include(${input_file})
   # copy user file to binary dir, just for 'visual' check.
   configure_file(${input_file} ${input_file} COPYONLY)
+  assert(WITH_FORTRAN "The option WITH_FORTRAN is not defined in the user options file")
 else()
```
xhub: Why do we need that?

control/control_tests.cmake
```diff
@@ -15,8 +15,11 @@ if(WITH_${COMPONENT}_TESTING)
   # Simulation tests
   BEGIN_TEST(src/tests)
   
-  NEW_TEST(tests PIDTest.cpp SMCTest.cpp ObserverTest.cpp)
+  NEW_TEST(tests)
   
+  IF(WITH_FORTRAN)
+    NEW_TEST(PIDTest.cpp SMCTest.cpp ObserverTest.cpp)
+  ENDIF()
   END_TEST()
```
xhub: This is a bit misleading: it should be WITH_EVENTDRIVEN. We only have fortran code that perform an event-driven integration. If we have SUNDIALS bindings, then we can use the latter for performing ED integration.

Ideally I would switch inside the test to a TS OSI if no ED OSI are available.

control/swig/CMakeLists.txt
```diff
@@ -23,4 +23,8 @@ set(${COMPONENT}_SWIG_DEFS "${SWIG_DEFS}")
 
 include(swig_python_tools)
 swig_module_setup(${COMPONENT}_PYTHON_MODULES)
-build_python_tests()
\ No newline at end of file
+if(NOT WITH_FORTRAN)
+  list(APPEND ${COMPONENT}_python_excluded_tests
+    ${CMAKE_CURRENT_SOURCE_DIR}/tests/test_smc.py)
+endif()
+build_python_tests()
```
xhub: ditto WITH_EVENTDRIVEN

kernel/kernel_tests.cmake
```diff
@@ -76,9 +76,12 @@ if(WITH_${COMPONENT}_TESTING)
...
 
-  NEW_TEST(testSimulationTools ZOHTest.cpp OSNSPTest.cpp)
-
+  NEW_TEST(testSimulationTools  OSNSPTest.cpp)
 
+  IF(WITH_FORTRAN)
+    NEW_TEST(ZOHTest.cpp)
+  ENDIF()
+  
   END_TEST()
```
xhub: ditto WITH_EVENTDRIVEN

kernel/swig/CMakeLists.txt
```diff
@@ -29,6 +29,11 @@ if(NOT WITH_XML)
   list(APPEND ${COMPONENT}_python_excluded_tests
     ${CMAKE_CURRENT_SOURCE_DIR}/tests/test_xml.py)
 endif()
+if(NOT WITH_FORTRAN)
+  list(APPEND ${COMPONENT}_python_excluded_tests
+    ${CMAKE_CURRENT_SOURCE_DIR}/tests/test_matrix_exp.py)
+endif()
+
 build_python_tests()
```
xhub: ditto WITH_EVENTDRIVEN

xhub (approved):
It's a good idea to have this flag. I would like to see a WITH_EVENTDRIVEN option to control whether some control tests are performed.

vacary:
I think with have to separate WITH_FORTRAN form WITH_ EVENTDRIVEN since we will implement sundials inregration

PR #204, opened 2018-01-31
green travis

bremond:
Profiling and serialization are out (==> Jenkins)
nix : from siconos.nix, only make is tested (no configuration)


vacary:
cool.

PR #205, opened 2018-01-31
Lazy initialization

vacary:
Just to start a discussion. Do not merge for the moment.


vacary:
Here is a first proposition to change a more lazy initialization process.

Remain to do:
- kernel : add a log of changes in nsds to initially only what has changed
- mechanics : 
   - be careful with the class SpaceFilter that appears to be an incomplete base class (@bremond could you have a look ?)
   - simulation->link has been changed to nsds->link. To be checked carefully that it works
   - mechanisms has to be updated
- control : the model has been replaced by nsds and simulation. @xhub could you check carefully that I have not mixed the plant nsds with the process nsds ?
- io : Nothing is done. 
   - Should we separate the serialization of the nsds and the simulation ?
   - is it possible to put a nsds and a simulation in single archive ?

vacary:
Remain to do:
- kernel :
   - add a log of changes in nsds to initialize only what has changed
   - review initialize_new in simulation to remove what has to be done only one time
- mechanics : 
   - be careful with the class SpaceFilter that appears to be an incomplete base class (@bremond could you have a look ?)
   - simulation->link has been changed to nsds->link. To be checked carefully that it works
   - mechanisms has to be updated
- control : the model has been replaced by nsds and simulation. @xhub could you check carefully that I have not mixed the plant nsds with the process nsds ?
- io : 
    - we serialize from the simulation

radarsat1:
I have found that the ref file for BulletBouncingBox.cpp does not match anymore in the latest lazy_initialization head.  It does match where I had it before, quite a few commits back in f41b911c9b13e529b4ba4c5b12c11edd2b96f190.  Meanwhile the ref file for BouncingBallTS.cpp matches just fine, so it seems something changed wrt to the collision engine.  Currently I am doing a bisect to find what changed.

radarsat1:
It seems this is due to moving `updateInteractions()` back outside the Newton loop.
I had put it in `TimeStepping::initializeNewtonLoop()`, between `computeInitialNewtonState` and `updateOutput`/`Input`.  To work better with lazy initialization I moved it to `Simulation::initialize()`, such that it's now before `computeInitialNewtonState`.

If I move it back to after `computeInitialNewtonState` then the ref files pass.

Now I am not sure which is more correct.  The idea of putting it after initializeNewtonLoop is that it creates the contact points after moving forward one step, and so their locations should be more accurate.  I have verified that it does tend to prevent more interpenetration.  On the other hand this could contribute to passing through thin surfaces for fast-moving objects.  I found that leaving it in both places still passes the ref file, which is nice.  This make sense in any case since Bullet uses persistent contacts, so setting up contacts at the beginning of the timestep and updating them after the initial move should be robust.  (With a bit more work we could defer removal of contacts until the next timestep..)

I think for now I will put back the call to `updateInteractions` after `computeInitialNewtonState`, and further discussion of robust/predictive contact detection can be left for another discussion.

vacary:
After this last commit, more all the examples are running well. There are still some examples that are broken,  but they were already broken on master

radarsat1:
I have been using this for a while and ironed out a few more issues. I will merge it since I don't think we will go back on this work.  The changelog is a very useful idea imho.  (E.g., I think we can use it for automatically tracking graph copies for multiple simulations, etc.)

PR #209, opened 2018-02-12
Lazy initialization wo ds link

vacary:
This PR and this branch is a continuation of the branch lazy_initialization. I made a new branch and a new PR to discuss the particular point of removing DSlink from the graph.

The goal was to be able to call a computeInput or ComputeOutput of an interaction before the simulation initialization and without invoking interactionProperties in the signature. This need is particularly important in the control class but also for a standalone use of the modeling class in Python.

It was not mandatory to move DSLink from the graph to the interaction but it render clearer the interface. would be possible to leave into the graph and to have a reference on it the Interaction class.

We should put in the graph the work and temporary variables needed for the simulation and we try to avoid to expose the graph to the user in the modeling part.

It can be discussed altogether,


radarsat1:
I think this could benefit from the development of an example of a Simulation-less use of Siconos in order to fully understand the needs of DSLink in this case.

radarsat1:
Now I can use the changeLog to detect whether there are new links, I no longer need to track `_linkOrUnlink` when calling `updateInteractions()` to know whether to call `initOSNS`.

Now, I wonder if `initializeInteraction` is necessary in `Simulation::link()`, and whether `initOSNS` is necessary in `Simulation::updateInteractions()`, since they could be called as necessary from `Simulation::initialize()`.  If so, then `Simulation::link()` and `Simulation::unlink()` become no-ops w.r.t to `NonSmoothDynamicalSystem::link()` / `NonSmoothDynamicalSystem::removeInteraction()`, and could be removed.

radarsat1:
Please notice my comment in `Simulation::updateInteractions`:
```C++
  if (_nsdsChangeLogPosition != _nsds->changeLogPosition()) {                            
    initOSNS();                                                                          
                                                                                         
    // Since initOSNS calls updateIndexSets() which resets the                           
    // topology->hasChanged() flag, it must be specified explicitly.                     
    // Otherwise OneStepNSProblem may fail to update its matrices.                       
    _nsds->topology()->setHasChanged(true);                                              
  }    
```

Now that I am looking at it again this seems very messy.  Can it be improved?  The problem is that when interactions are added or removed we call initOSNS which calls updateIndexSets which resets the Topology hasChanged flag.  However, hasChanged is used to tell whether to do anything in computeOneStepNSProblem, so we must reset it manually.

Perhaps the changeLog can replace the hasChanged flag completely.  I'll try to propose something..

radarsat1:
Ok, conclusion after working on it for a while; the `setHasChanged(true)` thing is okay, as it basically just undoes the effects of it being reset by updateIndexSets(), but is nonetheless needed later.  A similar solution is to call `setHasChanged()` from `OSNS::initialize()`, but this seems more general so I keep it.  However, I moved the initOSNS call to `Simulation::initialize()`, it no longer needs to be done during updateInteraction.  I move updateInteraction to `Simulation::initialize()` as well, so that it's outside the Newton loop, so that new interactions can be detected.  The idea of updating interactions inside the newton loop was never very sound, and it was broken in this new version of the initialization code, so after making a failed effort to fix it, I decided it's better to remove it.

vacary:
@radarsat1 I review your changes and it sounds great.  the fact that you remove linkOrUnlink simplifies again the initialization process. I think we can propose to remove also Simulation::link and Simulation::unlink at the end of the process.

I merge this PR into lazy_initialization.

PR #210, opened 2018-02-14
Lazy initialization with order

vacary:
I started a new branch from lazy_initialization.  The goal in this new branch is also to improve the interface of modeling for the end user. This is a work in progress.

Do not merge before we complete all the examples and we review the impact of such modifications

Before this branch, I was really difficult to understand why we can have access to some members of a DS class and some other not. The same for the interaction with computeOutput and computeInput methods that  generate SEGFAULT if these methods were called with the wrong derivative order or level.

For instance, It was possible to do 
```
# the dynamical system
ds = sk.LagrangianLinearTIDS(x, v, mass)
relation  = sk.LagrangianLinearTIR(H)
nslaw = sk.NewtonImpactNSL(e)
inter = sk.Interaction(nslaw, relation)
nsds = sk.NonSmoothDynamicalSystem(t0, tend)
nsds.insertDynamicalSystem(ds)
nsds.link(inter, ds)
ds.q()
ds.velocity()
inter.computeOutput(t0,0)
inter.computeOutput(t0,1)
inter.computeInput(t0,1)
```

but any of these calls generates a segfault or an assertion in debug mode

```
ds.acceleration()
inter.computeOutput(t0,2)
inter.computeInput(t0,0)
inter.computeInput(t0,2)
```
After the first step for an EventDriven simulation,  all these calls are working.

It comes from several aspects of the siconos interface:
- in dynamical systems, we allocate by default several levels for the state (_x, or _q) and this is not very clear from the interface
- the level of output in interaction may be different from the level of derivatives in DS. Generally, we
have  y[i] <-- f(_x[j], j<= i). So if the DS has _x[j] j<=i, y[i] should be allocated and accessible. In the words the storage order in DS should drive the level of derivative in interactions output
- Conversely, the levels for lambda in the interaction are set by default thanks to a visitor, but the non smooth input is not necessarily allocated. If lambda[0] exists, perhaps p[0] is not allocated.

With @fperignon. we propose to 
- add an "order" info in DS that corresponds to the derivative number of the state that is allocated in the DS and we perform the allocation at construction.  ds.acceleration() will return an Exception if the level is not sufficient.
- the output of the interaction is allocated accordingly to the order of DS.
- we use the input levels in interaction for setting non smooth output in DS. In particular, if no interaction, the non smooth input is not allocated.

One of the question is now:

Do we authorize the simulation  to change the order in DS and levels of input in Interaction (it is the case right now)? If we authorize,
- Pro:
   - nothing to change in the API 
   - more flexibility
- cons
   - the accessor and methods change their behavior after the first advanceToEvent() 
   - we have to work a lot to update DSlink and other vectors in Modeling part.
If we do not authorize:
- Pro:
   - the behavior of the model is more clear for the user.
   - the simulation does no change anything in the DSlink. (Only the work vectors and metrics are created.)
- cons
   - change in the API requiring the user to set the right order and levels of input such that the simulation will work. (if you want to use Lsodar, construct DS or order 2)


vacary:
some test of the new interfaces are there  https://github.com/siconos/siconos/pull/210/commits/394fc2a2a01c6ff9e4fa26c6d682ab586a9bf435#diff-562d2eb4cb70efaf1e167de9fee8457d

radarsat1:
I suggest maybe we merge lazy_initialization_wo_DSLink, or at least lazy_initialization, into master before proceeding too far?  It's getting a little confusing which branch to work on for other things.. (I am currently revising contact detection code, and was working in lazy_initialization_wo_DSLink.  I could move to master but fear needing a lot of rebasing/conflict resolution later.)

After using for a bit I am confident that the changelog method works quite well in practice.  It only requires an extra call to clear the log once in a while, but it simplifies a lot of things.  So far no problems encountered with the DSLink change as well, but I am not sure what the right call is for where it should live.

As for this PR, still thinking about it.  I see the pros and cons.  I think somehow it makes sense for the Simulation to verify or even set up the levels of the DS/Interactions, but I am not sure whether it should be an explicit step.  (Goes against the "lazy initialization" idea but it could easily be confusing if these things change after the user sets up the model.)

radarsat1:
okay I see you've merged lazy_initialization_wo_DSLink into lazy_initialization, so I guess the latter is the branch to work from for now.

vacary:
ok, I suggest that we merge lazy_initialization into master as soon as possible. Some examples are still failing due to slight differences with the ref as you mentioned.

vacary:
Just a follow-up, It think we can choose a go-between solution. We add the order attributes in DynamicalSystem, but the OSI can change it if needed. I will try to commit something in that sense

vacary:
I close it for the moment

Issue #156, opened 2017-06-20
Make it possible to run vview without Siconos or mechanics_io


radarsat1:
It should be possible to run vview without building/installing all of Siconos.

PR #223, opened 2018-03-12
[io] split mechanics_io into mechanics_hdf5 and mechanics_run.

radarsat1:
mechanics_hdf5 contains only the "write" parts of "mechanics_io", for
creation of HDF5 files.  mechanics_run inherits from it and adds the
"import" and "run" parts that require Siconos Mechanics.

With this change, vview, vexport, et al can run without Siconos compiled libraries
installed, as long as VTK and h5py is available.  One can also create and view
an HDF5 file locally without installing Siconos and copy it to another
computer that has Siconos installed to run the simulation.

Would fix issue #156 and enable creating a simple Siconos "viewer" app, as well as importers for various applications (e.g. Blender, ParaView..) that don't require Siconos to be installed locally.


vacary:
Nice work. I propose to merge.

PR #227, opened 2018-03-13
Refactor vview/vexport into a class

radarsat1:
This is some (imho) much needed work to refactor vview and vexport from a long script with many nested functions into a class VView() with organized functions.

The `siconos_vview` and `siconos_vexport` scripts now become stubs that basically just instantiate the VView class and call the appropriate functions.  In particular vexport instantiates all the same VTK objects but does not initialize the renderer windows, but calls VView.export() instead of VView.run().

I left vexport.py for now so that output can be compared, I still need to verify that nothing is missing in the output files.

Also, I understand that vexport.py is an attempt to *remove* vexport functionality from vview.py, while this puts it back in, and so may be seen as wasted time, but I saw that there was just so much repeated code there, that almost the entire differences between vview and vexport could be summarized in a single function, VView.export().  If it's undesired it could still be moved to a separate script and added using inheritance instead of reimplementing everything.

This follows a fairly common approach with Python tools of putting all the functionality in a module/class and putting only very minimal code in /usr/bin front-ends to import and call it.

You can run `pydoc siconos.io.vview` after installing to see how the class is organized.


vacary:
Thanks for this refactoring

PR #232, opened 2018-03-27
Mechanics occ fixes

bremond:
face edge contact issue fixed and joint with clearance added in occ_slider_crank.py
normalFromFace1 removed (normal is always on face2). offset2 added on second contact.


vacary:
Thanks

PR #237, opened 2018-04-05
Build without path and sort

radarsat1:
These two patches make the build work if I remove folders externals/PATH_SDK and externals/sort.  These are the two main license-related pain points for #236.  Please review.

### DIFF
--- a/numerics/src/tools/NumericsSparseMatrix.c
+++ b/numerics/src/tools/NumericsSparseMatrix.c
@@ -27,6 +27,7 @@
 #include "NumericsSparseMatrix.h"
 #include "numerics_verbose.h"
 #include "NumericsMatrix.h"
+#include "string.h" // memcpy
 
 #include "debug.h"
 
--- a/numerics/src/tools/NumericsSparseMatrix.c
+++ b/numerics/src/tools/NumericsSparseMatrix.c
@@ -45,10 +45,20 @@ typedef struct
   size_t indx;
 } sort_indices_struct;
 
+#ifdef HAVE_SORT
 #define SORT_NAME sorter
 #define SORT_TYPE sort_indices_struct
 #define SORT_CMP(x, y) ((x).i - (y).i)
 #include "sort.h"
+#else
+#include "stdlib.h" // qsort
+static int sort_indices_struct_cmp(const void *a, const void *b)
+{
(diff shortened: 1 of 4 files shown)


numerics/src/tools/NumericsSparseMatrix.c
```diff
@@ -27,6 +27,7 @@
 #include "NumericsSparseMatrix.h"
 #include "numerics_verbose.h"
 #include "NumericsMatrix.h"
+#include "string.h" // memcpy
```
xhub: shouldn't that be #include <string.h>

numerics/src/tools/NumericsSparseMatrix.c
```diff
@@ -44,10 +45,20 @@ typedef struct
...
 } sort_indices_struct;
 
+#ifdef HAVE_SORT
 #define SORT_NAME sorter
 #define SORT_TYPE sort_indices_struct
 #define SORT_CMP(x, y) ((x).i - (y).i)
 #include "sort.h"
+#else
+#include "stdlib.h" // qsort
+static int sort_indices_struct_cmp(const void *a, const void *b)
```
xhub: ditto

numerics/src/tools/NumericsSparseMatrix.c
```diff
@@ -44,10 +45,20 @@ typedef struct
...
+{
+  const sort_indices_struct *sa = (const sort_indices_struct *) a;
+  const sort_indices_struct *sb = (const sort_indices_struct *) b;
+  return (sa->i > sb->i) - (sa->i < sb->i);
+}
```
xhub: I was just curious about that magic: is sa->i - sb->i not sufficient? If my C99 is not too rusty, sa->i > sb->i evaluates to a boolean. So the result is a difference of boolean promoted to int. It is a bit confusing for me. I have to admit that I don't remember that piece of code very well ...

radarsat1:
Thanks, quite right about the brackets, will fix.  The sort comparison code is adapted straight from the documentation, the idea is that it's comparing two non-inclusive inequalities so if they're less it's -1, greater it's 1, and equal it's 0.  (like strcmp)  It should be correct I think.

radarsat1:
This documentation that is: https://www.gnu.org/software/libc/manual/html_node/Comparison-Functions.html#Comparison-Functions

PR #241, opened 2018-05-23
[kernel] remove some updates in initializeNewtonLoop

vacary:
The following calls are already made in simulation::initialize
```
updateWorldFromDS();	
updateInteractions();
updateContactsFromChangeLog();
```
I propose to remove it. What do you think ?


radarsat1:
The difference between the updateInteractions in Simulation::initialize and the one in TimeStepping::initializeNewtonLoop is before and after MoreauJeanOSI::computeInitialNewtonState which calls MoreauJeanOSI::updatePosition.  So the two calls do not update interactions based on the same state.

I experimentally found better behaviour taking both states into account during construction of interactions for the Newton loop, but correctness should be considered carefully.  Removing one of these will likely change some ref files.

vacary:
Ok, I propose that we close this PR without merging. The best we can do for the moment is to open an issue on this subject. see #276

PR #250, opened 2018-09-10
Nsm diag indices trivial

vacary:
I propose to simplify the routine 
```
CS_INT* NSM_diag_indices(NumericsMatrix* M)
```

The previous implementation is really expensive and causes huge slow down of the ADMM solvers. the new implementation in 
```
CS_INT* NSM_diag_indices_trivial(NumericsMatrix* M)
```
is a quite naive implementation, but more efficient.


xhub:
The function NSM_diag_indices_trivial and NSM_diag_indices are not equivalent.
NSM_diag_indices_trivial gives incorrect results when the sparse matrix is missing diagonal elements.
NSM_diag_indices is expensive becacuse it ensures that the matrix will have all diagonal elements. There is a bunch of memcpy to make sure that the matrix is replaced by one that has all diagonal elements. The copy of the matrix could be optimized.

I can only guess at this point, but I think that many matrices have all diagonal elements present in the representation. It would not suprise me that those coming from SBM have that properties, as well as most matrices from FC problems. I would rather have code that is guaranteed to always return the right answer (given some hypothesis on some properties).

I see a few ways forward:
- add a property (boolean) that the matrix possess all diagonal elements. In that case use the logic in NSM_diag_indices_trivial to find them.
- populate the diagonal indices when the matrix is created.
- try to find all the diagonal indices( ala NSM_diag_indices_trivial). If some are missing, add them (with the code as the current NSM_diag_indices). This could lead to some speedups.


Also, it is a bit orthogonal, but the name NM_add_to_diag3 is a bit misleasing, since in the NM_SPARSE and dense cases, it works on any matrices. It is just the SBM case that is restricted to 3x3 blocks.

vacary:
Thanks for the explanation.
I tried in 4bc6154 to implement your last option.

It is clear that NM_add_to_diag3 is a bit weird. Should we clarify it ?

numerics/src/FrictionContact/fc3d_admm.c
```diff
@@ -158,6 +159,7 @@ static void fc3d_admm_symmetric(FrictionContactProblem* restrict problem,
     DEBUG_PRINT("Force a copy to sparse storage type\n");
     M = NM_create(NM_SPARSE,  problem->M->size0,  problem->M->size1);
     NM_copy_to_sparse(problem->M, M);
+    NSM_diag_indices_trivial(M);
   }
```
xhub: Can this be NSM_diag_indices?

xhub (changes requested):
I left 3 comments (sorry should have done a proper review)

It's good to go if those are addressed in some way

numerics/src/tools/NumericsSparseMatrix.h
```diff
@@ -160,6 +160,9 @@ extern "C"
    */
   CS_INT* NSM_diag_indices(NumericsMatrix* M);
 
+  CS_INT* NSM_diag_indices_trivial(NumericsMatrix* M);
+
```
xhub: Can NSM_diag_indices_trivial be a static function inside NumericsSparseMatrix.c?

That way it not in the ``public API'' and doesn't get used?

vacary: I agree. This is done in the commit below

vacary:
@xhub All your comments have been taken into account

xhub:
I merge it. Note that a test fails with MUMPS

https://cdash-tripop.inrialpes.fr/testDetails.php?test=27441&build=931

PR #252, opened 2018-09-11
[externals] use sort distribution from github https://github.com/swenson

vacary:
use sort distribution from github https://github.com/swenson/sort

	    commit  d360ba02ca55803673ead2497d352135a9aa291e
	    local modifications:
	      + mv c test files in test
	      + add missing prototypes
	      + fix warning on comparisons between differents types

This version has a clear licence. I was obliged to correct some missing prototypes and some warning  on comparisons.

The good way is perhaps to propose a pull request in https://github.com/swenson/sort
To be discussed.


xhub (approved):
LGTM

That version of sort.h looks maintained too.

PR #255, opened 2018-09-26
[io] siconos_vexport: add --global-filter: one vtp file / time step

bremond:
This is only a quick fix for the filesystem saturation issue. It just appends a vtkCompositeDataGeometryFilter before the writer (which becomes a simple vtkXMLPolyDataWriter instead of a vtkXMLMultiBlockDataWriter). A single vtp file is written for each time step instead of a directory with a file for each objects. Note that as everything is merged in one big polydata the object id information is lost. Warnings from vtkDataSetSurfaceFilter indicates that all added arrays are also lost (contact info, velocities). vtkCompositeDataGeometryFilter is also very slow.

### DIFF
--- a/io/swig/io/vview.py
+++ b/io/swig/io/vview.py
@@ -249,16 +249,17 @@ class VExportOptions(VViewOptions):
         if long:
             print()
             print("""Options:
-            --help     display this message
-            --version  display version information
-            --ascii    export file in ascii format
+            --help           display this message
+            --version        display version information
+            --global-filter  one vtp file/time step
+            --ascii          export file in ascii format
             """)
 
     def parse(self):
         ## Parse command line
         try:
             opts, args = getopt.gnu_getopt(sys.argv[1:], '',
-                                           ['help','version','ascii'])
+                                           ['help','version','ascii', 'global-filter'])
             self.configure(opts, args)
         except getopt.GetoptError as err:
                 sys.stderr.write('{0}\n'.format(str(err)))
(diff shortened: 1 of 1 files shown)


vacary:
Could you just give how we can use it ?
Thanks

bremond:
for example:

siconos_vexport --global-filter chute_con_rocas-uniform-mu-2.0.hdf5

should output something like this:

 [...]
  -rw-rw-r--   1 maurice maurice 2.7M Sep 26 11:38 chute_con_rocas-uniform-mu-2.0-7.vtp
  -rw-rw-r--   1 maurice maurice 2.7M Sep 26 11:38 chute_con_rocas-uniform-mu-2.0-6.vtp
  -rw-rw-r--   1 maurice maurice 2.7M Sep 26 11:38 chute_con_rocas-uniform-mu-2.0-5.vtp
  -rw-rw-r--   1 maurice maurice 2.7M Sep 26 11:38 chute_con_rocas-uniform-mu-2.0-4.vtp
  -rw-rw-r--   1 maurice maurice 2.7M Sep 26 11:38 chute_con_rocas-uniform-mu-2.0-3.vtp
  -rw-rw-r--   1 maurice maurice 2.7M Sep 26 11:38 chute_con_rocas-uniform-mu-2.0-2.vtp
  -rw-rw-r--   1 maurice maurice 2.7M Sep 26 11:38 chute_con_rocas-uniform-mu-2.0-1.vtp
  -rw-rw-r--   1 maurice maurice 2.7M Sep 26 11:38 chute_con_rocas-uniform-mu-2.0-0.vtp

PR #258, opened 2018-10-22
[cmake] fix MUMPS detection on Suse (and debian?)

xhub:
This PR would fix building with MUMPS on Suse and possibly Debian.

Reviewers: please check and test if possible.


radarsat1:
This completely fixes the IDONTWANTMPI problem I was having on Debian.  Thank you!
If no one sees any problems I would like to merge it..

bremond:
I made an opensuse ci config for travis

https://github.com/bremond/siconos/tree/ci-add-opensuse

The build in progress is here:
https://travis-ci.org/bremond/siconos/builds/445047454

But how do you install mumps on opensuse?

I'am stuck on

862c0dfa0974:/ #.(B zypper search mumps
Loading repository data...
Reading installed packages...
No matching items found.

xhub:
Thanks a lot for looking into this.
 
I don't have neither root nor sudo access so I can't test the commands. The documentation says

zypper ar -f <URL>

where URL is the url for the repo that has the distro version in it: see the list in https://build.opensuse.org/repositories/science

I don't know what the "latest" opensuse will be, so I can't tell. That's not even how it is install on my work machine ... the trvis logs are not helpful. According to https://hub.docker.com/r/library/opensuse/, latest is 42.3 ,so the URL should be http://download.opensuse.org/repositories/science/openSUSE_Leap_42.3/

bremond:
Ok I managed to add the science repo to install the sequential mumps on opensuse,
the configuration is working with your fix.

PR #267, opened 2018-11-14
[io] mechanics_run: fix #263 with option 3, without warning

### LINKED ISSUE #263: About inertia of objects with Bullet Contactors
With python mechanics_io and Bullet contactors, if inertia was not given at the add_object level, we had an implicit computation of inertia, due to the fact that python object None was mapped to a null SimpleMatrix shared pointer.
This implicit computation is done at each time step (updateContactorInertia, with bullet ), from the fact that BodyDS.setUseContactorInertia is set to true at the construction if inertia is not given.
Now, from c6ec73960, the mapping is not done anymore (we discussed this with @vacary and we agree that these kind of type mapping, None to null pointer, is very, very dangerous as it may occur in situation where it is not wanted). So this implicit inertia computation is not done and the simulation from master branch ends at the moment with a failure in those cases, see for example cube.py.
What solution is preferable:
- leave it like this, and force the user to provide an inertia matrix. Note that this breaks the compatibility and this means to rewrite the examples. the setUseContactorInertia becomes useless (correct me if I am wrong).
Also this means to provide an API hook to calculate inertia before add_object. For occ the inertia is implictly computed if some Volume objects are specified at the same place of the Contactor objects, but what can be done directly with Bullet is not as general as what can be done with OpenCascade.
- if the inertia is not given, print some warning about the implicit calculation and compute the inertia before the BodyDS creation by the same bullet call as the one in updateContactorInertia. Again this means that setUseContactorInertia becomes useless.
- if the inertia is not given, print some warning, build a null shared pointer on SimpleMatrix and pass it at BodyDS creation (or find another way to get the same result). In this case, we get nearly what we had before, but we know why.
- another solution ?

bremond:
A fix proposal with third option of #263, without warning: we call the BodyDS constructor without inertia and we set the flag etUseContactorInertia to True. Note that second option would be more complicate as we do not have access anymore to bullet contactors from python, they are associated to SiconosShape at the construction of the ContactManager. With the ContactManager framework the correct way to do this would be to implement ourself each inertia computation.

### DIFF
--- a/io/swig/io/mechanics_run.py
+++ b/io/swig/io/mechanics_run.py
@@ -909,23 +909,22 @@ class MechanicsHdf5Runner(siconos.io.mechanics_hdf5.MechanicsHdf5):
                 }
 
             else:
-                # a moving object
 
-                if inertia is not None:
-                    if np.shape(inertia) == (3,):
+                if inertia is not None and np.shape(inertia) == (3,):
                         inertia = np.diag(inertia)
-                    elif np.shape(inertia) != (3,3):
-                        print('Wrong shape of inertia')
-                    have_inertia = True
-                else:
-                    inertia=0.0*np.eye(3) # set a temporary null value
-                    have_inertia = False
 
-                body = body_class(translation + orientation,
-                                  velocity,
-                                  mass, inertia)
-                if have_inertia:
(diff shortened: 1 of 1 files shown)


vacary:
fine by me

vacary:
I propose to merge.

PR #273, opened 2018-12-14
Second order ds

vacary:
The object of this PR is to add a SecondOrderDS class as a base class for NewtonEulerDS and LagrangianDS, It should:

- help to mix both class in examples
- factorize some code in OSI and OSNS


xhub (changes requested):
I like the idea. I just noticed some potential issues with virtual (or-non virtual) methods

PR #275, opened 2018-12-26
Simplify numerics tests

vacary:
Hi,
Just to start to reduce the targets for the tests in numerics in order to avoid to long outputs.
Not perfect but works.

I will continue for LCP and GFC3D, and so on.

Comments are welcome if you think that it may be more simplified. My goal was to avoir the use of python.


vacary:
The number of target for the tests has been divided by 4 and we run now more tests that before :-)

The time of building on Travis has also severely reduced.

@xhub For the moment, I am able to test GAMS/PATH solvers. If you have time ?

bremond:
Some details I can see:
1. the commit e65f1beb (mechanics_hdf5.py) does not seem to be related to the numerics tests simplifications,
2. the numerics tests output is far too verbose (cf ADMM iterations output) and the visible log is cut by travis now,
3. some tests are failings:
 on opensuse 4.3 with mumps_seq
         58 - fc3d_test_nsn_collection_1 (Failed)

	 59 - fc3d_test_nsn_collection_2 (Failed)

	 60 - fc3d_test_nsn_collection_3 (Failed)

	 84 - ConvexQP_test_collection (Failed)

        (on this distrib 2 other tests are also failing:
         24 - python_test_sparse, (Failed)
         96 - python_test_matrix_exp, (Failed) but this is the same on the master branch)
note that on master we have:
        24 - python_test_sparse, (Failed)
	427 - fc3d__NSN_AC_Tol_1e-5_Max_1000_inTol_0_inMax_0_IPARAM_1_Capsules-i100-1090 (Failed)
	952 - python_test_matrix_exp, (Failed)

 on ubuntu 16.04 with_mumps_par:
      ConvexQP_test_collection (Failed)

vacary:
The first commit is already also in master. You can forget it.
I remove the verbose in ADMM solvers.
I set a solver as unstable with MUMPS. Mumps generates a lot of instability curiously with Lapack. This has to be checked carefully

For opensuse+mumps, the master branch is also not working. I remove it for a while from the default TRAVIS builds

PR #284, opened 2019-03-24
Rolling friction

vacary:
Implementation of rolling friction.

Two points must be reviewed carefully that prevent for the moment to merge.
1- in MechanicsIO.cpp, I was not able to put to ANSWER macro in the ForMu() method
2- in mechanics_run.py the choice of RollingFrictionContact in the run method is  little bit weird. This has to be rationalized


vacary:
I tried to 
- improve the management of multiple nonsnooth laws in mechanics_run. This is not perfect yet since the constructor of NonSmooth law for joints in done implicitlely
- correct the code following the recommendation of codacy.

I propose to merge and to work on templates classes in a further branch

PR #288, opened 2019-03-27
[io] Fix path and target for serialization headers generation.

radarsat1:
After some mentions of build problems I thought I would check the status of master.  I am unable to build with serialization turned on.  I also found that generating the serialization headers is broken.

This PR changes the `doxygen` dependency target which has changed I guess.  Probably `kernel_xml4swig` is not correct, maybe someone can suggest a better fix.

However, I think the patch to `build_common.py` is necessary.

Unfortunately even with this, I get errors in `kernelPYTHON_wrap.cxx` trying to build with serialization, but I cannot figure them out.

Many errors like this:

```
/usr/include/boost/serialization/access.hpp:116:11: error: ‘class LinearOSNS’ has no member named ‘serialize’; did you mean ‘serializable’?
```

which seems like the classes do not have serialization declared but it's weird because `SiconosFullGenerated.hpp` is definitely getting included, I used `#error` to check.  However I noticed that the generated headers have a lot of things re-ordered, which is supposed to be avoided by sorting them by priority and by name in `io/tools/build_from_doxygen.py`, but it seems to have decided to move a lot of things around so it's hard to tell if this is due to registration order.

### DIFF
--- a/io/CMakeLists.txt
+++ b/io/CMakeLists.txt
@@ -92,7 +92,7 @@ if(WITH_SERIALIZATION)
 
     ADD_CUSTOM_COMMAND(
       OUTPUT ${GENERATED_HEADER}
-      DEPENDS doxygen ${GENERATED_DEPENDENCIES}
+      DEPENDS kernel_xml4swig ${GENERATED_DEPENDENCIES}
       ${CMAKE_SOURCE_DIR}/io/tools/build_from_doxygen.py
       ${CMAKE_SOURCE_DIR}/io/tools/builder_common.py
       ${GENERATED_COMMAND})
@@ -104,7 +104,7 @@ if(WITH_SERIALIZATION)
 
     ADD_CUSTOM_COMMAND(
       OUTPUT ${GENERATED_HEADER_KERNEL}
-      DEPENDS doxygen ${GENERATED_DEPENDENCIES}
+      DEPENDS kernel_xml4swig ${GENERATED_DEPENDENCIES}
       ${GENERATED_COMMAND_KERNEL})
   endif()
 endif()
--- a/io/tools/build_from_doxygen.py
+++ b/io/tools/build_from_doxygen.py
@@ -47,16 +47,16 @@ def get_classes_conditional(doxy_xml_files, cond):
                     })
     return found
 
-def classes_from_build_path(build_path):
+def classes_from_build_path(build_path, targets):
     """Get classes and members from all Doxygen XML files found on the
        provided build path."""
-    doxy_xml_path = os.path.join(build_path,'Docs/build/html/doxygen/xml')
+    doxy_xml_path = os.path.join(build_path,'docs/build/doxygen/doxy2swig-xml')
     if not os.path.exists(doxy_xml_path):
         print('%s: Error, path "%s" does not exist.'%(sys.argv[0], doxy_xml_path))
         sys.exit(1)
-    doxy_xml_files = (
-        glob(os.path.join(doxy_xml_path, 'class*.xml'))
-        + glob(os.path.join(doxy_xml_path, 'struct*.xml')))
+    doxy_xml_files = [p for t in targets
+                      for p in (glob(os.path.join(doxy_xml_path, '%s/class*.xml'%t))
+                                + glob(os.path.join(doxy_xml_path, '%s/struct*.xml'%t)))]
 
     # We want only classes that contain calls to the
     # ACCEPT_SERIALIZATION macro.
@@ -142,8 +142,7 @@ if __name__=='__main__':
      build_path) = parse_args(need_build_path=True)
 
     all_headers = get_headers(targets)
-
-    doxygen_classes = classes_from_build_path(build_path)
+    doxygen_classes = classes_from_build_path(build_path, targets)
 
     header_classes = classes_from_headers(all_headers, include_paths)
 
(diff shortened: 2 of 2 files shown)


radarsat1:
I should mention that if I turn off serialization things seem to build fine.

radarsat1:
Further info:

Without this patch, the errors are in new objects SiconosCone, SiconosCapsule, they do not have a "serialize" method.  I suppose they could be added manually, but the SerializersGeneration should work.

Then, with `-DWITH_GENERATION=ON`, then `make SerializersGeneration` leads to,

```
make[3]: *** No rule to make target '../io/doxygen', needed by '../io/src/SiconosFullGenerated.hpp'.  Stop.
```

So it is related to recent changes in docs.

With this patch now it depends on xml doxygen targets, and I have updated (rebase) the PR to look for xml in `doxy2swig-xml/{components}/*.xml`, and this works for generation, but I still get a lot of errors in `kernelPYTHON_wrap.cxx` when trying to compile.

So, currently, I don't know how to fully fix this.

But I think it's clear that there must be added a CI target that includes serialization and is preceded by SerializersGeneration, if this feature is to remain maintainable.

radarsat1:
>  do we need python docstrings from swig to run serialization

Not sure, I don't think so.

> (which is implied if kernel_xml4swig is a dep)

Yes, I just used it to ensure xml was generated.  Perhaps we need an xml4serialization target, but not sure why multiple "xml" targets are needed, there is just one set of xml outputs from doxygen, is this not the case?  It seems there are now multiple directories under `Docs/build/xml`, but I am not up to date on how you are rearranging things.

> And I guess that we should be able to run serialization/generation even when PYTHON_WRAPPER is off ?

Agreed.  Recall that the doxygen xml was just used as a replacement for gccxml.  Previously I think I always used SerializersGeneration with WITH_DOCUMENTATION=1 so I am not 100% confident that the dependency chain is perfect.  It would be good for it to depend explicitly on doxygen xml generation.

> do we need xml files for all components ? Only kernel ? Some others ? (I would say all ?)

It is used to generate `SiconosFullKernelGenerated.hpp` which should only depend on kernel, and `SiconosFullGenerated.hpp` which additionally includes classes from mechanics and control.

There is a dict called `input_headers` that maps these components to the headers they need the xml for in `io/tools/builder_common.py`.

The way it works is that `classes_from_build_path` in `build_from_doxygen.py` opens all xml files and scans for classes that support serialization.  These are compared against a list of classes to ignore.  The remaining classes are ordered according to priority and then alphabetically.  These are then written to the `SiconosGenerated*.hpp` headers along with all members and superclass dependencies.

xhub:
Just a quick comment that I had the same thoughts as @fperignon about the dependency on kernel_xml4swig, and the fact that this would not work when the python bindings are not enabled.

I got lost in trying to come up with a good proposal.

@radarsat1 your work on that is very much appreciated.

radarsat1:
ps

> Another remark : if I understood properly the add_custom_command(s) in io/CMakelist.txt, headers (GENERATED_HEADER and GENERATED_HEADER_KERNEL) are generated into source dir rather than into binary dir. Isn't it a bit strange ?

Yes I always found that strange but never fixed it.  Principally I think it is because we do not include this as part of the build process, but rather we generate these files and then commit them as part of the source.  We could always change it to be a standard part of the build process, the only effect would be to require doxygen every time we build with serialization, but I do not think this is too bad a requirement.

By the way, with regards to CI, I found that turning on serialization increases requirements for building the python bindings so much that building kernel.i requires 8 GB of memory, which has the effect of crashing with out-of-memory errors in at least the debian salsa CI system, as well as my personal laptop.  (I have to close Firefox to compile Siconos successfully!)  So be careful.

xhub:
The memory requirements are a big issues. It has to be documented: I had the same issue 4 years ago. The fix would be to split that gigantic file. However, IIRC, a hack is needed to have all the .o compiled into 1 python module (.so).

fperignon:
"Perhaps we need an xml4serialization target, but not sure why multiple "xml" targets are needed, there is just one set of xml outputs from doxygen, is this not the case? It seems there are now multiple directories under Docs/build/xml, but I am not up to date on how you are rearranging things."


yes there are multiple xml outputs and targets.
I tried to explain thing here : https://nonsmooth.gricad-pages.univ-grenoble-alpes.fr/siconos/devel_guide/write_and_build_doc.html <https://nonsmooth.gricad-pages.univ-grenoble-alpes.fr/siconos/devel_guide/write_and_build_doc.html>
and through comments in docs/gendoctools files.

To sum it up :

A first pipeline leads from headers/xml outputs to rst files for sphinx  to handle python/docstrings in swig 
- hpp —> xml —> .i (doxy2swig) —> rst .
More or less what is done in gendoctools/python2rst.py

Another pipeline leads from xml outputs  to rst files for sphinx to handle c++ comments :
- hpp —> xml —> rst (breathe complient) 
More or less what is done in gendoctools/cpp2rst.py

And to make things easy the doxygen config to generate xml files can not be the same for swig and breathe.

So we need different targets :

<component>_xml4swig  (For python API in autodoc)  —> files in <cmake_binary_dir>/docs/build/doxygen/doxy2swig-xml/<component>

<component>-doxy2xml (For C++ API in autodoc)   —> files in <cmake_binary_dir>/docs/build/doxygen/xml4rst/<component>

So, taking that into account and regarding Steve comments, I think a dependency to kernel_doxy2xml will be enough

I have commit something to fix this, or at least try to …

PR #300, opened 2019-09-11
[mechanics] scale inertia to worldscale.

bremond:
In mechanics_run python interface, when inertia is not given it is computed by bullet (or occ).
When worldscale is not 1, the inertia computed by bullet is left unchanged which is not correct.
I propose this scaling ```(1./(woldscale**2))```

For ```example spheres_in_a_box.py``` I checked with 

```python
mass = (4./3)*pi*(radii[i]**3)*2320
I_ = (2./5)*mass*(radii[i]*radii[i])
print('python inertia:',[I_, I_, I_])```

and then traces in ```SiconosBulletCollisionManager.cpp```

```python inertia: [2.968471848146946e-10, 2.968471848146946e-10, 2.968471848146946e-10]
sphere radius:2.38004
calculated inertia for ds:2, with mass:0.000131013 :2.96853e-10 2.96853e-10 2.96853e-10
```

The loss of precision is probably due to single float usage in the bullet library I use (the one from debian)


vacary:
It is ok for me.
Just a question: how Bullet compute the inertia from the mass ? Bullet deduces the density of the material from the mass and the volume ?

vacary:
@bremond: Should we scale inertia to world scale or to the square or the world scale.

vacary:
@bremond: Should we scale inertia to world scale or to the square or the world scale.

bremond:
In this pull request, the scale factor for inertia was computed  in `mechanics/src/collision/bullet/SiconosBulletCollisionManager.cpp`. line 745.  It was `scale_factor = 1./(_options.worldScale*_options.worldScale);` and was applied during `updateContactorInertia`, so with this patch inertia should not be scaled by the user before.

In the master branch, ` _options.worldScale` is used for the scale of shapes(*)  but is not used for inertia. This PR tried to improve this.

(*) mesh shapes are not scaled, note this TODO:   `// TODO: worldScale on vertices` for `SiconosMesh` line 1261

PR #309, opened 2019-10-28
Cholesky Factorization

vacary:


numerics/src/tools/NumericsMatrix.c
```diff
@@ -2891,6 +2891,200 @@ int NM_gesv_expert(NumericsMatrix* A, double *b, unsigned keep)
...
+      {
+        /* the mumps instance is initialized (call with job=-1) */
+        NM_MUMPS_set_control_params(A);
+        NM_MUMPS(A, -1);
+        NM_MUMPS_set_verbosity(A, verbose);
+        NM_MUMPS_set_icntl(A, 24, 1); // Null pivot row detection
+        NM_MUMPS_set_cntl(A, 5, 1.e20); // Fixation, recommended value
+      }
+      NM_MUMPS_set_problem(A, b);
+
```
xhub: I think we should call `NM_MUMPS_set_sym(A, 2)` here to set the proper symmetry value? In `NM_MUMPS_set_control_params` the `sym` value is set to `0`. In the MUMPS manual, `sym` can have 3 values: `0` for (unsymmetric) general matrices, `1` for symmetric positive definite, `2` for general symmetric.

vacary: Yes I left the same code than for NM_gesv. This has to be done. 
I was waiting to discuss with @bremond to know it we need another interface for LDLT or if we can put in NM_posv. 
For interior point method we need is explicitly  and not a simple Cholesky 
Thanks for the comment

vacary:
I close this PR that has been partially merged. 
I added an issue #312 to take into account the remark of @xhub

vacary:
Start to do the job. If you're interested in :

https://github.com/vacary/siconos/tree/LUandCo

Issue #296, opened 2019-06-20
Other kind of matrices for MassMatrix in BouncingBallTS


bremond:
If one tries to build a banded matrix in the BouncingBall:

```c++
SP::SiconosMatrix Mass(new SimpleMatrix(nDof, nDof, Siconos::BANDED));
```
(which would be something as a tremendous optimization...),
this fails.

The problem is in ```SimpleMatrixSolvers.cpp:50```
 ```c++
void SimpleMatrix::PLUFactorizationInPlace()
```

where the case Siconos::BANDED is not checked, only the dense and sparse cases are checked.

Moreover, in the sparse case only the symmetric case is treated
```
 int info = cholesky_decompose(*sparse());
    // \warning: VA 24/11/2010: work only for symmetric matrices. Should be replaced by efficient implementatation (e.g. mumps )
``

vacary:
More generally, we should improve the simpleMatrix to provide more linear solvers.

vacary:
The #319 PR should help to take into account

PR #319, opened 2019-12-04
LU and co

vacary:
Start to work on on :

- reformatting the fonctions NM_gesv, NM_posv in numerics
- using a NumericsMatrix for solving linear systems in kernel


vacary:
@bremond I start to add MUMPS support. Since I am not sure about the whole process of using MUMPS, could you review my last commit.

On a other subject, I think that NM_KEEP_FACTORS is no longer needed.

bremond:
Hello,
I added the possibility to specify multiple dense rhs for the mumps solver in the last commit, so just one solve is needed in this case, a test is made in NM_MUMPS_test.c for symmetric and unsymmetric matrices.
The rhs was a vector, it is now a matrix. The number of columns of the rhs matrix is set with a nrhs parameter in N_MUMPS_set_problem. The number of rows of the rhs matrix is set to the number of rows of the factorized matrix, but it can also be superior to it. I don't know if it can be of any usage here but one have to keep this in mind.
The rhs matrix can also be sparse, but this would need quite more refactoring, so let's think about it later. 
I agree that NM_KEEP_FACTORS should be definitively removed, but this is also the case for NM_NONE, NM_PRESERVE.
For the dense case only, there is a flag isLUfactorized, I think we should generalize this flag to 
the sparse case, so NM_factorize should do nothing if it is true, and it should be the responsibility of the caller to ensure the flag is unset if a factorization is needed. 
I propose to start the transformation of all NM_gesv to NM_LU_factorize + NM_LU_solve, if you do not have already done it ?

vacary:
Thank you very much for your work.

1- I didn't made the global change from NM_gesv to NM_LU_factorize + NM_LU_solve. You can do if you have time.

2- Let us remove   NM_KEEP_FACTORS and the others. The only interesting one is NM_PRESERVE, but we can assume that the user may manage a copy of its data.  We can use a Boolean of type _isFactorized for all actorizations. but we have to define some rules:
    - should we change the Boolean when we do a copy into a matrix ?
    - when we call directly NM_LU_factorize should we use this Boolean or to force to factorization ?
   - if we use NM_LU_solve, then we factorize only _isFactorized is false.

3- With @fperignon , we have started to correct to correct/update all the examples 
https://my.cdash.org/index.php?project=Siconos+Examples&date=2020-03-16
for the new release 4.3.0. It could be good if we can merge this PR, but since it is a bit important form the kernel side, I would prefer to pass the examples.

vacary:
We need also to take into account the remark in #296

bremond:
So I pushed a draft where the usage of the flags `NM_KEEP_FACTORS`, 
`NM_PRESERVE` (and `NM_NONE`) is suppressed in `NM_LU_solve` and
 `NM_LU_factorize`.

For factors (was `NM_KEEP_FACTORS` flag):

the default behaviour is to factorize once. A new factorization is 
only triggered if:

 - the matrix `isFactorized` flag is false which is the case at
   initialisation,

 - the user sets this flag explicitely with `NM_set_factorized(A,
   flag)`.

For the matrix preservation (was `NM_PRESERVE`):

The matrix owns a pointer called `destructible` which points to the
matrix itself by default. A call to `NM_preserve(A)` makes this pointer
points on a copy of the matrix. `NM_factorize` and `NM_LU_solve` works
always on `A->destructible` so they do not have to take care of
preservation or not: this reduces the code size and complexity.
`NM_factorized`, `NM_set_factorized` works also on this pointer.

`NM_preserve(A)` makes a copy only if `A->destructible` points on `A`.  If
one wants to force the copy, a call to `NM_unpreserve(A)` is necessary
before : it clears the copy if it exists.

Numerics tests are ok. The old calls to `NM_gesv_expert` are kept in comments.

On the kernel side it is very weird for me for the moment.

vacary:
Thank you very much. For me, is is nearly perfect.
I have now to correct all the bugs that I have introduced in the Kernel.

bremond:
Note that code for `UMFPACK`, `SUPERLU`, `PARDISO` are not inserted in NM_LU_factorize. Probably it is not a lot of more than a cut and paste, but it must be checked. An I did not even checked with MUMPS yet, I ran the tests only for `lu_sol`. 
Also the symmetric case is stil in the old flavor (`NM_posv_expert`).

vacary:
Yes, this is the reason why I prefer to merge this PR after the release 4.3.0, that should arrive soon.

vacary:
With the last commit, all the tests and the examples are working well.

The goal of this PR was to improve the linear solvers for sparse matrices. Unfortunately, we switch from the Boost-like Cholesky solvers to the LU factorizations for all. This means that :
- we need to continue our effort to implement NM_Cholesky_* and NM_LDLT_* in numerics to be able to call solvers for symmetric matrices
- we need also to have a default behaviors in the Kernel to choose the most efficient factorization if the matrices are symmetric.

vacary:
On the side Kernel, the Cholesky factorization can be now called by the Kernel. For that, the user must specify that the matrix is symmetric and definite positive. On the Impacting bar example, we save time with the Cholesky factorization, but there is still some issues:
 - The computation of the block of the OneStepNSproblem is not able to work with Sparse matrices. The operation `getLeftInteractionBlock` especially constructs a dense matrix
- Futhermore, solving with sparse rhs is not implemented in Kernel. It seems that it is possible with MUMPS but not with cholesky_solve from cparse.

If we want to deal with large FE mesh, it can be very expensive for the cpu time.

vacary:
For the first point, `getLeftInteractionBlock` and `getRightInteractionBlock` keep the type of the original Matrix specified by the user

vacary:
From a comment of T.A. Davis here https://github.com/JuliaLang/julia/issues/33637, it seems possible to solve system with Sparse r.h.s. using cs_spsolve  . We have to discuss how to do it in Siconos. Perhaps, it could be good to have the following signature 👎 
```
NM_LU_solve(NumericsMatrix * A, double b) // Solve Ax=b for a single dense vector
NM_LU_solve_matrix_rhs(NumericsMatrix * A, NumericsMatrix * B) // Solve A X=B for a generic matrix (possibly sparse)
```

@ bremond What do you think about that ?

vacary:
@bremond all the tests and the example are ok. I propose to merge this branch and to restart a new PR for the question of solving linear system with sparse r.h.s.

What do you think ?

PR #318, opened 2019-12-03
Symmetric mumps

bremond:
This adds MUMPS solver possibility to `
NM_posv_expert` with `half_triplet` matrix as a new storage.

Tested in :
NM_MUMPS_test, NM_test, gfc3d_TEST_ADMM_COLLECTION_3

PR #320, opened 2019-12-04
Some valgrind fixes for gfc3d admm (work in progress)

bremond:
On big problems (such as `spheres-in-a-box-39998-i10000-115560-13.hdf5`one can see the memory increase with a simple `top` command.
This a work in progress as there is still leaks, for example with the usage of `NM_transpose`, cf `Htrans.` I still do not understand why it is not freed.

Many errors come from the fact that `NM_free` does not free the pointer! I Propose to rename
`NM_free` and `NSM_free` to `NM_clear` and `NSM_clear.`


bremond:
I close this PR, as the changes are on top of #318, an I open a new one directly on top of master history

PR #321, opened 2019-12-06
Some valgrind fixes for numerics and gfc3d admm

bremond:
This comes from some valgrind traces on gfc3d admm tests.

Many errors come from the fact that NM_free does not free the pointer! I Propose in this PR, along with a few fixes in gfc3d admm, to rename NM_free and NSM_free to NM_clear and NSM_clear.
But It seems like there is a lot of other places in numerics code where free is not done after these *_clear calls.


vacary:
ok, thank you. I merge it on gricad side

PR #327, opened 2020-03-24
[numerics,kernel] use of nullptr rather than NULL to comply with C++1…

vacary:
…1 standard


vacary:
It seems that #fperignon did the job in //.  We will forget this commit after merge of @fperignon

Issue #328, opened 2020-03-25
compile error with serialization


radarsat1:
Here are the complete flags I am trying to compile with (debian unstable):

```
$ cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=None -DCMAKE_INSTALL_SYSCONFDIR=/etc -DCMAKE_INSTALL_LOCALSTATEDIR=/var -DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON -DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=ON -DCMAKE_INSTALL_RUNSTATEDIR=/run "-GUnix Makefiles" -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_AUTOGEN_VERBOSE=ON -DCMAKE_INSTALL_LIBDIR=lib/x86_64-linux-gnu -DCMAKE_SKIP_BUILD_RPATH=TRUE -DFORCE_SKIP_RPATH=1 -DCMAKE_SKIP_INSTALL_RPATH=TRUE -DCMAKE_SKIP_RPATH=TRUE -DINSTALL_CMAKE_DIR=/usr/lib/x86_64-linux-gnu/cmake/siconos -DPYTHON_EXECUTABLE=/usr/bin/python3 -DI_WANT_STATIC_LPSOLVE=ON -DLpSolve_LIBRARY=/usr/lib/liblpsolve55_pic.a -DWITH_TESTING=ON -DWITH_DOCUMENTATION=ON -DWITH_DOXY2SWIG=ON -DWITH_SYSTEM_SUITESPARSE=ON -DWITH_SERIALIZATION=ON -DWITH_BULLET=ON -DWITH_FCLIB=OFF -DWITH_OCE=OFF -DIDONTWANTMPI=ON -DWITH_MUMPS=ON ..
```

The compilation ends with,
```
/home/sinclairs/siconos/bld/wrap/siconos/numericsPYTHON_wrap.cxx:3570:10: fatal error: SiconosFullNumerics.hpp: No such file or directory
 3570 | #include <SiconosFullNumerics.hpp>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

radarsat1:
The path io/numerics/src is _not_ in the gcc command, not sure why:

```
[...]

PR #329, opened 2020-03-25
Fix serializersgeneration

vacary:


vacary:
Is there a kink with #328 ?

PR #331, opened 2020-03-28
Fix serializers generation

radarsat1:
Please check the commits, it's mostly path fixes that can maybe be solved a better way!

Also, bear in mind that this enables ENABLE_PREPROCESSING in Doxygen, let's see if it doesn't mess up the docs in any significant way.

(Not enabling preprocessing was causing parsing errors with e.g. `TYPEDEF_SPTR(..)` leading to classes not being documented.)


radarsat1:
Unfortunately I found that compilation still fails with serialization when `WITH_BULLET=ON`.

I'm having trouble fixing some problems at the boundary of Siconos and Bullet classes.

radarsat1:
For example there are problems with the constructor and member variable that takes a btManifoldPoint in Bullet1DR.  I'm having trouble tracing why this was not an issue with BulletFrom1DLocalFrameR.

Otherwise there seem to be errors related to shared pointers in the swig modules, so maybe some registrations missing in `.i` files, but I'm not sure what the problem is.

The error messages are quite useless unfortunately, e.g.,

```
/home/sinclairs/siconos/build/wrap/siconos/mechanics/collision/basePYTHON_wrap.cxx:7837:37:   required from here
/usr/include/boost/serialization/shared_ptr.hpp:269:5: error: non-constant condition for static assertion
```

vacary:
@radarsat1 Thank you for the job.

For Bullet1DR, I change a bit he interface when I added other types of relations. Perhaps, I make a mistake. I saw also that the class Contact2d2D is considered as incomplete by swig. This has to be reviewed carefully.

for the shared pointers, no idea ..

I propose to merge your PR to see how react the CI, and to reopen it after to continue to work on this subject.

radarsat1:
> I propose to merge your PR to see how react the CI

yes that's okay, up until I added WITH_BULLET, it was actually working XD...

PR #343, opened 2020-05-05
Matrix versioning

bremond:
This is only a draft of numerics matrix versioning following what we said on gitter. Versions are initialized at creation, a few assertion are made on the versions of NSM->origin or NM->storageType. All tests are ok. It is far from replacing the current mechanism now, but, if after some thinking it appears it is worth going in this direction, it may be a base for its improvement.
I tried to keep the low level versioning isolated in a `NumericsDataVersioning.h` file  (the version is only a number at the moment, but, for example, for debugging sha1 signatures could be added).  
A few version management functions are introduced In `NumericsSparseMatrix.h` and `NumericsMatrix.h`.


vacary:
@bremond It seems that there is a missing file for the compilation of this branch.

``` src/tools/NumericsMatrix.h:31:10: fatal error: NumericsDataVersion.h: No such file or directory ```

see https://my.cdash.org/viewBuildError.php?buildid=1884800

vacary:
@bremond Do you think that we can merge ?

bremond:
I think it can be merged as it it is very short and isolated, but note
that, at this stage, it brings only a few assertions and the system
remains the same.  The only reason to merge is to prevent some breaking
by further evolutions in numerics, but for such short code this is not
very important. There is also a «convert to draft» button, on the PR
page on github, maybe I can try ?

vacary:
Ok, I am not able to merge directly. Could you have a look to choose the right code and merge ?

PR #353, opened 2020-10-15
[numerics] add MUMPS for sparse rhs (LU & Cholesky)

bremond:
This is the MUMPS part for the sparse rhs case (LU & Cholesky). Note that the result is a dense matrix and it is copied from the mumps_id struct to the input/output B matrix. Only the MUMPS centralized configuration (i.e full matrices on host 0) is used.


vacary:
Merged !

PR #354, opened 2020-10-15
LDL support with SuiteSparse

vacary:
In this PR, I use LDL solver from SuiteSparse. If SuiteSparse is installed on the system, we use it. Otherwise, we do exactly as for CXSparse. We extract a subset of routines of SuiteSparse copied in externals/SuiteSparse and we build siconos with it

@bremond The way we choose the solver by default is a bit problematic. 
https://github.com/siconos/siconos/blob/master/numerics/src/tools/NumericsSparseMatrix.c line 295

If I switch the MA57 is MA57 is present, then all the other solvers (NM_LU_* NM_Cholesky_*) are no longer working

Perhaps, we can fallback to NSM_CSPARSE, even if the solver is set to NSM_HSL, if the LU or Cholesky is not implemented in NSM_HSL ?

I do not know if we can do better ?


bremond:
If MA57 is meant to be used only in LDLT case, or just waiting  for the implementation in Cholesky and LU cases, let's have another flag specific for this case.
Something like a check in LDLT_solve on (p->LDLT_solver) instead of p->solver with default to p->solver.

PR #368, opened 2020-11-20
[kernel] implementation of checkCompatibleNSLaw  in OSNSP

vacary:


kernel/src/simulationTools/LinearOSNS.hpp
```diff
@@ -266,6 +266,8 @@ class LinearOSNS : public OneStepNSProblem
     _keepLambdaAndYState = val ;
   }
 
+  virtual bool checkCompatibleNSLaw(NonSmoothLaw& nslaw) {return true;};
+
```
xhub, line 270: nitpick: why not a pure virtual so that a child class must implement this method?

xhub:
The function `checkCompatibleNSLaw` has not been added to AVI, Relay, MLCP and others. Is it on worth doing it for all (and once and for all)?

kernel/src/simulationTools/LinearOSNS.hpp
vacary: agree. Just I am lazy and I have not finished to implement all the classes. I will fix that at the end :-)

xhub: Sounds good :)

vacary:
@xhub  Thanks for your review. I completed to PR in my last commit following your recommendation

vacary:
Do not hesitate to check that I didn't forget some compatible nonsmooth laws.

xhub:
I think that GeneralMechanical is missing, but this may be done in the other PR?

There is not "virtual" or "override" keyword added to `checkCompatibleNSLaw` in the child classes. My C++ is too rusty to remember if it does really matter. I don't think we expect to have child class of those. That my usual nitpicking :)

vacary:
> I think that GeneralMechanical is missing, but this may be done in the other PR?

Yes, I added the method that do nothing

> There is not "virtual" or "override" keyword added to `checkCompatibleNSLaw` in the child classes. My C++ is too rusty to remember if it does really matter. I don't think we expect to have child class of those. That my usual nitpicking :)

For the moment, there is no second-level child classes, to I think it is sufficient. no ?

xhub:
> > There is not "virtual" or "override" keyword added to `checkCompatibleNSLaw` in the child classes. My C++ is too rusty to remember if it does really matter. I don't think we expect to have child class of those. That my usual nitpicking :)
> 
> For the moment, there is no second-level child classes, to I think it is sufficient. no ?

Yes, I was nitpicking as usual :)

PR #407, opened 2021-06-01
Simplify interaction

vacary:
Remove yOld lamdaOld and y_k from Interaction


fperignon (approved):
Thanks, I don't even understand why these 'old' stuffs were still here doing the same job as y_k and lambdaMemory.

PR #434, opened 2021-12-17
[numerics] add a default case for N(S)M_version_copy

bremond:
This prevents some undefined behavior and avoids warnings about the switch statements in the 2 new functions.

### DIFF
--- a/numerics/src/tools/NumericsMatrix.c
+++ b/numerics/src/tools/NumericsMatrix.c
@@ -2745,7 +2745,11 @@ void NM_version_copy(const NumericsMatrix* const A, NumericsMatrix* B)
     NSM_version_copy(A->matrix2, B->matrix2);
     break;
   }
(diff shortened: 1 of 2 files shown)


vacary:
I have still  a compilation error:
```
/Users/acary/siconos/numerics/src/tools/NumericsSparseMatrix.c:278:5: error: void function 'NSM_version_copy' should not return a value [-Wreturn-type]
    return 0;
    ^      ~
1 error generated.
make[2]: *** [numerics/CMakeFiles/numerics.dir/src/tools/NumericsSparseMatrix.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
/Users/acary/siconos/numerics/src/tools/NumericsMatrix.c:2750:5: error: void function 'NM_version_copy' should not return a value [-Wreturn-type]
    return 0;
```

PR #435, opened 2021-12-17
Io output staticbody number

vacary:


vacary:
@bremond I would like to have your point of   view on this PR? There is less members in the BulletR* relations classes in that way, but I need to use a map to link the pointer on the body shape record to its smart pointer. I do not know if it is a good idea ?

bremond:
Hy, I think that, in SiconosBulletCollisionManager, keeping track of another copy of the smart pointer is useless, as a copy is already inserted in bodyShapeMap (so the data won't go away). When you need a smart pointer from a raw pointer you can use createSprClassName as for example createSPtrbtManifoldPoint in the same file. Here you have to remove the const qualifier as bodyShapeRecordA are non const members :
```
   rel->bodyShapeRecordA = createSPtrBodyBulletShapeRecord(*const_cast<BodyBulletShapeRecord*>(pairA));
   rel->bodyShapeRecordB = createSPtrBodyBulletShapeRecord(*const_cast<BodyBulletShapeRecord*>(pairB));
```

(I need also to add DEFINE_SPTR(BodyBulletShapeRecord) in BulletSiconosFwd.hpp). This compiles, I didn't check anything.

(The createSptr* functions are defined in SiconosPointers.hpp. They symply associate an empty deleter  with a smart pointer).

PR #442, opened 2022-02-10
[io] h5py: direct check of attributes instead of version

bremond:
This  should fix https://github.com/siconos/siconos/runs/4559708594?check_suite_focus=true

which is due to a wrong decision made on h5py version.
With python, it is safer to check the existence of attributes with ``hasattr``.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -38,6 +38,7 @@ jobs:
         liblpsolve55-dev \
         python3 \
         python3-packaging \
+        python3-h5py \
         libpython3-dev \
         libcppunit-dev \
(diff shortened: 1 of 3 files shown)


bremond:
It is ok with the latest commit : https://github.com/bremond/siconos/runs/5160565165?check_suite_focus=true
To summary : the first commit is necessary on my machine (ubuntu 21.04, python 3.9), but I am not sure if it is needed on the workflow docker (ubuntu 20.04, python 3.8). The second commit adds the possibility to have the native collision detection without an installation of bullet (which is the case on the workflow docker). The last commit adds the missing h5py package to the workflow docker.

PR #443, opened 2022-02-24
Fix size in fclib interface

vacary:
1) fix sizes in fclib interfaces for triplet matrices
	   We output only values up to nz and we set nzmax to nz
 	   There  is no interest to save in fclib file values from nz to nzmax
2) use NM_triplet by default for writing problem in fclib_local format
	   (simplify code)

### DIFF
--- a/numerics/src/FrictionContact/fclib_interface.c
+++ b/numerics/src/FrictionContact/fclib_interface.c
@@ -273,12 +273,8 @@ int frictionContact_fclib_write(FrictionContactProblem* problem, char * title, c
   fclib_problem->info->description = description;
   fclib_problem->info->math_info = mathInfo;
 
-  fclib_problem->W = (struct fclib_matrix*)malloc(sizeof(struct fclib_matrix));
   fclib_problem->R = NULL;
   fclib_problem->V = NULL;
-
-  fclib_problem->W->m = problem->M->size0;
-  fclib_problem->W->n = problem->M->size1;
 
   CSparseMatrix * spmat = NM_triplet(problem->M);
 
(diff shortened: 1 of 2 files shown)


fperignon (approved):

numerics/src/FrictionContact/fclib_interface.c
```diff
@@ -250,6 +252,65 @@ int frictionContact_fclib_write(FrictionContactProblem* problem, char * title, c
...
+  fclib_problem->W = (struct fclib_matrix*)malloc(sizeof(struct fclib_matrix));
+  fclib_problem->R = NULL;
+  fclib_problem->V = NULL;
+
+  fclib_problem->W->m = problem->M->size0;
+  fclib_problem->W->n = problem->M->size1;
+
+  CSparseMatrix * spmat = NM_triplet(problem->M);
+
+  fclib_problem->W = (struct fclib_matrix*)malloc(sizeof(struct fclib_matrix));
```
bremond, line 285: fclib_problem->W is allocated a second time here (first time at line 276)

bremond (commented):
Just a detail about allocations, in frictionContact_fclib_write fclib_problem, fclib_info (fclib_problem.info) and fclib_matrix (fclib_problem.W) can be created directly on the stack without malloc (they are only structures). Only fclib_problem.W.p and fclib_problem.W.i need allocation on the heap (the size depends on the argument and may be huge). Also, it seems that fclib_problem->W is allocated twice.

numerics/src/FrictionContact/fclib_interface.c
vacary: Thank you. An ugly copy-paste.

PR #444, opened 2022-03-01
Sliding triangle mesh ch2d

vacary:
Try to implement a strategy to avoid contact with vertices when an heightmap in 2d is described by a set of "glued" convex hulls

### DIFF
--- a/io/swig/io/mechanics_hdf5.py
+++ b/io/swig/io/mechanics_hdf5.py
@@ -906,9 +906,18 @@ class MechanicsHdf5(object):
 
     def add_convex_shape(self, name, points,
                          insideMargin=None, outsideMargin=None,
-                         avoid_internal_edge_contact=False):
         """
         Add a convex shape defined by a list of points.
+        
+        outsideMargin is the value of margin that substract from the actual contact distance
+
+        If insideMargin is positive, the convex hull is shrunken by that amount 
+        (each face is moved by "shrink" length units towards the center along its normal).
+        This value is then added to outsideMargin to compensate the shrink. The convex hull
+        appears for the user as in its original size.
+
+        
+
         """
         # infer the dimension of the problem
         if np.shape(points)[1] == 2:
@@ -938,6 +947,9 @@ class MechanicsHdf5(object):
                             insideMargin=None, outsideMargin=None):
         """
         Add a primitive shape.
+
+        Todo: Comments on insideMargin and outsideMargin that depends on the
+              primitive shape type.
         """
         # infer the dimension of the problem
         if primitive == 'Disk' or primitive == 'Box2d' or primitive == 'Line' \
--- a/io/swig/io/mechanics_hdf5.py
+++ b/io/swig/io/mechanics_hdf5.py
@@ -905,7 +905,8 @@ class MechanicsHdf5(object):
             self._number_of_permanent_interactions += 1
 
     def add_convex_shape(self, name, points,
-                         insideMargin=None, outsideMargin=None):
+                         insideMargin=None, outsideMargin=None,
+                         avoid_internal_edge_contact=False):
         """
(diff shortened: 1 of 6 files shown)


mechanics/src/collision/bullet/SiconosBulletCollisionManager.cpp
```diff
@@ -1803,64 +1822,119 @@ void SiconosBulletCollisionManager_impl::createCollisionObject(
   // btConvexHullShape* childShape1 = new btConvexHullShape(&pts[0],rows, sizeof(btScalar)*3);
 
   btConvexHullShape * btch;
-  btch = new btConvexHullShape(&pts[0], rows, sizeof(btScalar)*3);  // Warning: Possible loss of memory since we cannot SP here
+  btch = new btConvexHullShape(&pts[0], rows, sizeof(btScalar)*3);  // Warning: Possible loss of memory
```
bremond, line 1825: It depends on what removeCollsionObject do on the bullet side, btConvex2d does not free its parameter. A  simple solution here is to create the pointer as a smart pointer (SP::btCOnvexHullShape is defined) and attach this smart pointer into a structure which has the same lifetime as the object (a bit like bodyShapeRecord).

mechanics/src/collision/bullet/SiconosBulletCollisionManager.cpp
```diff
@@ -1803,64 +1822,119 @@ void SiconosBulletCollisionManager_impl::createCollisionObject(
...
-  DEBUG_PRINTF("ch2d->insideMargin() = %8.4e\t, ch2d->outsideMargin() = %8.4e\n", ch2d->insideMargin(), ch2d->outsideMargin());
-  btch->setMargin((ch2d->insideMargin() + ch2d->outsideMargin()) * _options.worldScale);
+  DEBUG_EXPR(display_info_btConvexHullShape(*btch););
+
+  // recalc bounding box
   btch->recalcLocalAabb();
 
+
   SP::btConvex2dShape btconvex2d(new btConvex2dShape(btch));
+
```
bremond, line 1895: So if btch is a SP:: this would be btConvex2dShape(*btch) here

io/swig/io/mechanics_hdf5.py
```diff
@@ -929,6 +930,8 @@ def add_convex_shape(self, name, points,
             shape[:] = points[:]
             shape.attrs['type'] = 'convex'
             shape.attrs['id'] = self._number_of_shapes
+            if avoid_internal_edge_contact:
+                shape.attrs['avoid_internal_edge_contact'] = True
```
bremond, line 943: A detail, this can be written ```shape.attrs['avoid_internal_edge_contact'] = avoid_internal_edge_contact:```

mechanics/src/collision/SiconosShape.hpp
```diff
@@ -505,9 +505,20 @@ class SiconosConvexHull2d : public SiconosShape,
   ACCEPT_SERIALIZATION(SiconosConvexHull2d);
   SP::SiconosMatrix _vertices;
 
+  /* boolean to use the normal to the selected edge of the convexhull
+     to avoid contact with vertex */
+  bool _avoidInternalEdgeContact;
+
+
 public:
+  /* index of the first point of the selected to compute the normal edge (default=0) */
```
bremond, line 514: of the selected edge

io/swig/io/mechanics_hdf5.py
vacary: Initially, I wanted to have the possibility to test if the attributes are existing or not.

mechanics/src/collision/SiconosShape.hpp
vacary: I will correct it.

PR #448, opened 2022-04-06
[io] correctly initialize the object counter if the hdf5 file contains already objects

vacary:
This PR enables to add again some objects in another "with" use.


bremond (commented):
I have tested it, it works and it seems ok.  A detail, to avoid the unnecessary binding of ```name``` in the list comprehension you can use the ```values``` method of the h5 group:
```python
type_obj = [obj.attrs['type'] for  obj in self._input.values()]
```

Issue #450, opened 2022-05-13
Your opinion about a structure change in mechanics_run.py (run function).


batzkass:
Hi,
I have two propositions to submit about the way the `run()` function in mechanics_run.py works. I need both of them to make a better integration of Siconos inside PlatRock. But before making the actual code I just want to get your opinion about them to know if there is any chance that they will be included in Siconos main branch.

1. I don't know -- a priori -- the total number of steps I would need to complete a single rockfall. So I set an insanely high simulation time (`T`), and I suggest that the simulation end can be triggered with the return value of the hooks. It would give something like:
```
if end_run_iteration_hook is not None:
    hook_ret_value = self.log(end_run_iteration_hook.call, with_timer)(k)
    if(hook_ret_value==False):
        break
```
I can confirm that this works.

2. For some reasons, I need to launch the [main while loop](https://github.com/siconos/siconos/blob/740c370a4b430ec71436a34a8edb582533f49cf7/io/swig/io/mechanics_run.py#L3046) multiple times, without re-doing all the scene creation/loading above. I suggest to move all the code that currently resides in this `while` statement to another function, and call this function once from the current `run` function instead. I didn't tried it for now as its a bit painful, a bunch of local variables have to be copied/imported/moved somehow in the new function.

Would you agree with the above changes ? If yes, for the point (2.), do you know a clean trick to access the local variables from the new function ?

batzkass:
Well, just to say I'm currently working on an slightly different approach that should not require any changes into siconos mechanics_run.

vacary:
Hi,

For the first item, I fully agree with your proposition to be able to stop the 'while loop' over the time--steps using the output of the end_run_iteration_hook. This is a nice idea that modifies only very slightly the code. 

For the second point, we can split the run function into an initialization phase that import the model and set-up the simulation objects. I agree also with that but is a bit of work. Looking at the code, I think there is not so much variables to be pass as arguments. I can a look to that if you want.

batzkass:
I found an other way of doing things ... and it's almost clean :-D. The only hack is that I still needed to run the "init phase" without the while loop, but setting `T` to `0` blocks [here](https://github.com/siconos/siconos/blob/master/io/swig/io/mechanics_run.py#L2762). So I basically used the start run hook to replace the simulation `hasNextEvent()` method to a `return False` function :

```
 class Siconos_Start_Run_Hook():
	def __init__(self, platrock_sim):
		pass

	def initialize(self, io):
		self._io = io
[...]

PR #451, opened 2022-05-19
Improvements about [io] interfacing with other codes.

batzkass:
See issue #450 about PlatRock.
Note: it would be nice if one can set `T=None` to allow infinite siconos calculation, to be used in combination with the new "hook abort" feature. But I didn't do it to avoid changes in Siconos kernel.


vacary:
Coooool.

I will run the siconos tests on it this night, and then merge if everything is ok.

Best

Issue #460, opened 2022-09-09
Bug in ADMM solver


radarsat1:
I can only reproduce this when the package (current master on github, 629130d32c80647e2535e13f00840a322bd91447) is built with `CMAKE_BUILD_TYPE=Debug`, but I've verified that in release mode the problem is still present but the assertion is just skipped.

I get the following errors running "make test" in the `numerics` directory:

```
96% tests passed, 3 tests failed out of 79

Total Test time (real) = 315.64 sec

The following tests FAILED:
	 36 - fc3d_TEST_NSN_COLLECTION_1 (Failed)
	 37 - fc3d_TEST_NSN_COLLECTION_2 (Subprocess aborted)
	 64 - ConvexQP_test_collection (Subprocess aborted)

```

If I run `ConvexQP_test_collection` manually, I get the following output on `test_2`:

```
start test #2
test step 1
z[0]=-5.000000	PX[0]=0.000000
z[1]=-4.000000	PX[1]=0.000000
z[2]=-3.000000	PX[2]=0.000000
z[3]=-2.000000	PX[3]=0.000000
z[4]=-1.000000	PX[4]=0.000000
z[5]=0.000000	PX[5]=0.000000
z[6]=1.000000	PX[6]=0.000000
z[7]=2.000000	PX[7]=-5487458222656572253044242706512533009332565296704281842353464341423459222659694957657785623185792771358609158748433686054963230167620949827376493029768065102014423289909563218546376801907227961512137003210872266266533210566056108843254404124141251251786124500068864635864510729375034248770336586551263232.000000
z[8]=3.000000	PX[8]=0.000000
z[9]=4.000000	PX[9]=0.000000
q[0]=-1.000000	q[1]=-2.000000	q[2]=-3.000000	q[3]=-4.000000	q[4]=-5.000000	q[5]=-6.000000	q[6]=-7.000000	q[7]=-8.000000	q[8]=-9.000000	q[9]=-10.000000	test step 1
ConvexQP_test_collection: /home/sinclairs/projects/siconos/numerics/src/tools/NumericsMatrix.c:2993: NM_triplet: Assertion `NM_max_version(A) == NSM_version(A->matrix2, NSM_TRIPLET)' failed.
Aborted (core dumped)
```

The first problem is that PX is obviously invalid.  This can easily be fixed by adding,

```
cqp.m=10
```

[...]

PR #396, opened 2021-03-10
[numerics] NM storageType + NSM origin sync with matrix version

bremond:
This adds a synchronization between storageType (DENSE & SPARSE BLOCK), origin (SPARSE) and matrix version
with a call to ```NM_version_sync(M)```. The call is added at the end of every function that modify some storage at the NM_* level.

As different matrices may have the last version, there is an order specified in ```NM_latest_id``` and ```NSM_latest_id```:

  NM_DENSE > NM_SPARSE_BLOCK > NM_SPARSE
  and
  NSM_TRIPLET > NSM_HALF_TRIPLET > NSM_CSC > NSM_CSR

Some missing initialization in GenericMechanical solver are added and ```NM_triplet```, ```NM_half_triplet```, ```NM_csc```, ```NM_csr``` create the matrix again if the version is not the latest one (the creation was done only if the data is null).

PR #461, opened 2022-09-12
[numerics] add some version check & clear on top of some NM_*

bremond:
This is a part of #396. When NM_{triplet,half_triplet,CSC,CSR} are called a check on the version is done. If it is lower than
the maximum version stored in the matrix, then the storage is cleared. In #460 the triplet` matrix version has a lower version at the end of ``NM_TRIPLET`` which means that is not up to date.


radarsat1:
I can confirm that this seems to fix #460 !

PR #500, opened 2024-06-18
[numerics] NSM_latest_id + few others out of compilation unit

bremond:
NSM_latest_id and some other functions about sparse matrix versioning inside numerics appear to be time consuming during computeOneStep for the vkernel (here 2d_sphere_recirculation.py, performance=True, 20000 disks, step ~ 7000 of 10000 timesteps, compactNsearch on gpu, gcc-12)
Screenshot from 2024-06-17 12-26-07

The solution to move them directly in the header file and force the compiler to inline them provides an interesting gain  (time without inlining / time with inlining is around 1.3 in this case and maybe also interesting in the case of bullet simulations) .
On my PC, without inlining:
```bash
./vrun --env-ok --time 'duration: %E' src/siconos/tests/2d_sphere_recirculation.py vnative > 2d_sphere_recirculation_vnative_gpu.log
duration: 29:06.32
```
The same command with inlining gives something like:
```bash
duration: 22:03.04 
```

I did not test with static alone (i.e. without inline keyword), to see if the compiler choose a correct solution by itself, but as the involved code is small it is probably not important. 

This PR is also pushed on gricad : https://gricad-gitlab.univ-grenoble-alpes.fr/bremonma/siconos/-/commits/numerics-inlining

### DIFF
--- a/numerics/src/tools/NumericsSparseMatrix.c
+++ b/numerics/src/tools/NumericsSparseMatrix.c
@@ -68,10 +68,6 @@ static int sort_indices_struct_cmp(const void *a, const void *b)
 #pragma GCC diagnostic pop
 #endif
 
-version_t NSM_version(const NumericsSparseMatrix* M, NSM_t type)
-{
-  return NDV_value(&(M->versions[type]));
-}
 
 void NSM_set_version(NumericsSparseMatrix* M, NSM_t type,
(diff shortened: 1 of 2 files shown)


vacary:
Thank you for this PR. I will merge it asap.

